PR #80, opened 2020-09-28
feat(elements|ino-icon): add question mark icon

P1NHE4D:
✨ **Features:**
* Added a question mark icon to the `ino-icon` component

🎟️ **Resolves:** [ELEMENTS-679](https://jira.inovex.de/browse/ELEMENTS-679)


janivo (changes requested):
You have to run `yarn icon:integrate-icons` inside of the elements package to copy the icons to the storybook as well. Because right now the icon cannot be found if you try to look it up inside the storybook.

janivo (approved):

PR #82, opened 2020-09-28
refactor(elements): move from local font files to font package

### LINKED ISSUE #71: Use shared font files
In our packages (`elements` and `storybook`) we use custom fonts, which we host ourselves in our project. Currently the fonts are defined in the elements package (see https://github.com/inovex/elements/blob/master/packages/elements/src/components/styles/fonts.scss). The storybook package tries to import these fonts into its own package:
https://github.com/inovex/elements/blob/6af692b05baa80357ea2febf24632d8110fba706/packages/storybook/src/core/global.scss#L3

Since the import of the fonts is a relative import, the files cannot be found by the Storybook. Here we should find a way to store the fonts statically and use them in both packages.

### LINKED ISSUE #75: angular/react: Fonts can't be found
### Describe the bug
After updating to version 1.0.2, the selfhosted fonts cannot be found in the consumer project.

### To Reproduce
1. Add `@inovex.de/elements(-react/-angular)@1.0.2` package to your dependencies
2. Start the app
3. Open dev console in browser

### Expected behavior
No errors and the correct fonts to be used.

### Screenshots
React:
image
Angular:
image

### Your Environment
 

| Software | Version |
| -------- | ------- |
| @inovex.de/elements | 1.0.0 |
| Browser | Chrome 85.0.4183.102 |

### Possible Solution
We probably have to copy the fonts to the angular and react wrapper as well.

janivo:
closes #75
closes #71

⚙️ **Changes**:
- Removed local font files
- Add fonts as dependency
- Add script to require fonts on load (stencil-config -> `globalScript`)


packages/elements/stencil.config.ts
```diff
@@ -4,6 +4,7 @@ import { sass } from '@stencil/sass';
 export const config: Config = {
   buildEs5: false, // see https://github.com/ionic-team/stencil/issues/253
   globalStyle: './src/components/styles/variables.scss',
+  globalScript: './src/util/import-fonts.ts',
```
janivo, line 7: There still seems to be a problem with angular not executing the script `globalScript` on load. It's working correctly for the react and storybook package.

janivo: I will wait for #85 and try again with the newest version.

janivo:
⚙️ Changes:
- Add Function in core packages which get called by the angular wrapper to import the fonts

I tried to find another way but this seems the only one working right now. I'm not an angular expert at all but I tried many other solutions which did not quite work. This should do the trick for now. I tested it in our react and angular examples projects.

packages/elements/stencil.config.ts
janivo: Found another solution

janivo:
also closes #71

P1NHE4D (approved):

PR #83, opened 2020-09-28
feat(elements|ino-snackbar): add property to set custom timeout

### LINKED ISSUE #81: multiple ino-snackbar instances appear upon changing the value of a story knob
### Describe the bug
Storybook generates multiple instances of the `ino-snackbar` component upon changing the value of a single knob.

### To Reproduce
1. Go to https://elements.inovex.de/dist/latest/storybook/?path=/story/notification-ino-snackbar--default-usage
2. Open the footer containing the knobs by pressing `D`
3. Change the value of any knob
4. Click on the corresponding `ino-button` to open the snackbar
5. Multiple instances should appear on the screen even though the button has been pressed only once

### Expected behavior
Only a single instance of the `ino-snackbar` should appear if the button is pressed once.

### Screenshots
Screenshot 2020-09-28 at 17 32 03

### Your Environment
 

| Software | Version |
| -------- | ------- |
| @inovex.de/elements | 1.0.2 |
| Browser | Safari |
| Operating System | Mac OS |

P1NHE4D:
✨ **Features:**
* Added a new `inoTimeout` property to the `ino-snackbar` component that can be used to modify the time it takes until the snackbar disappears

🎟️ **Resolves:** [ELEMENTS-699](https://jira.inovex.de/browse/ELEMENTS-699)

**Note:** Does not resolve `#81`

### DIFF
--- a/packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
+++ b/packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
@@ -53,7 +53,9 @@ export class Snackbar implements ComponentInterface {
     );
     this.snackbarInstance.open();
     this.snackbarInstance.timeoutMs = -1;
-    setTimeout(() => this.snackbarInstance.close(), this.inoTimeout);
+    if (this.inoTimeout >= 0) {
+      setTimeout(() => this.snackbarInstance.close(), this.inoTimeout);
+    }
   }
 
   componentWillUnload() {
--- a/packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
+++ b/packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
@@ -51,11 +51,8 @@ export class Snackbar implements ComponentInterface {
     this.snackbarElement.addEventListener('MDCSnackbar:closing', e =>
       this.handleSnackbarHide(e)
     );
+    this.configureTimeout();
     this.snackbarInstance.open();
-    this.snackbarInstance.timeoutMs = -1;
-    if (this.inoTimeout >= 0) {
-      setTimeout(() => this.snackbarInstance.close(), this.inoTimeout);
-    }
   }
 
   componentWillUnload() {
@@ -68,6 +65,13 @@ export class Snackbar implements ComponentInterface {
   private handleSnackbarHide(e) {
     this.hideEl!.emit(true);
     e.stopPropagation();
+  }
+
+  private configureTimeout() {
+    this.snackbarInstance.timeoutMs = -1;
+    if (this.inoTimeout >= 0) {
+      setTimeout(() => this.snackbarInstance.close(), this.inoTimeout);
+    }
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
```diff
@@ -46,6 +52,10 @@ export class Snackbar implements ComponentInterface {
       this.handleSnackbarHide(e)
     );
     this.snackbarInstance.open();
+    this.snackbarInstance.timeoutMs = -1;
+    if (this.inoTimeout >= 0) {
+      setTimeout(() => this.snackbarInstance.close(), this.inoTimeout);
```
P1NHE4D, line 57: Regarding the use of the `setTimeout` function: According to the [mdc documentation](https://github.com/material-components/material-components-web/tree/master/packages/mdc-snackbar), the `timeoutMs` property only supports values between 4000 and 10000 ms.

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
```diff
@@ -46,6 +52,10 @@ export class Snackbar implements ComponentInterface {
       this.handleSnackbarHide(e)
     );
     this.snackbarInstance.open();
+    this.snackbarInstance.timeoutMs = -1;
```
janivo, line 55: I don't quite get the thoughts behind lines 55 to 57. Can't we just set the `this.snackbarInstance.timeoutMs` to `inoTimeout` and then call the `open` function? Otherwise, we rewrite the same logic of the MDC component, don't we?

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
janivo: What happens if the consumer provides a value outside that range? If there is no error or warning, we should throw one ourselves and add this information to the description of the property.

janivo (changes requested):

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
P1NHE4D: According to the documentation, an error will be thrown. However, I think we should allow the consumer to set the property to any value and not just to a value within a specific range.

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
P1NHE4D: As I mentioned in the comment below, `setTimeout` only supports values in a specific range. Hence, I disabled mdc's timeout mechanism and added the `setTimeout` function to allow the consumer to set the property to any value.

janivo: Alright, that's fine. 
One last thing: Coul you extract the timeout logic into its own function? For me, the `componentDidMount` hook is for setting up stuff like listeners and calling other functions, but not for the logic itself.

P1NHE4D: Will do 🙂

janivo (approved):

PR #116, opened 2020-10-06
feat(elements|ino-snackbar): add property for warning and error states

### LINKED ISSUE #115: Add color schemes to ino-snackbar
The user should be able to change the overall color scheme of the `ino-snackbar` component. Hence, I would like to add a new `inoColorScheme` property -  just like in the `ino-button` component - which can be set to either of the following color schemes: `primary`, `secondary`, `warning`, `error`

P1NHE4D:
✨ **Features:**
* Added a new `inoColorScheme` property to the `ino-snackbar` component

**Closes #115**


janivo:
The new color schemes do not seem to match the designer's specifications. In the specification, it's only the border color that changes with each scheme.

P1NHE4D:
> The new color schemes do not seem to match the designer's specifications. In the specification, it's only the border color that changes with each scheme.

whoops I overlooked the specifications - I'll fix it

packages/storybook/src/stories/ino-snackbar/ino-snackbar.stories.js
```diff
@@ -53,6 +53,7 @@ export const DefaultUsage = () => /*html*/ `
                 ino-message="${text('ino-message', sampleText)}"
                 ino-action-text="${text('ino-action-text', 'Anlegen')}"
                 ino-timeout="${number('ino-timeout', 5000)}"
+                ino-type="${select('ino-type', ['', 'error', 'warning'], '')}"
```
janivo, line 56: Examples in the storybook for `warning` and `error` would be nice

janivo (changes requested):

packages/storybook/src/stories/ino-snackbar/ino-snackbar.stories.js
P1NHE4D: ✅

packages/elements/src/components/ino-segment-button/readme.md
```diff
@@ -67,18 +67,6 @@ class MyComponent extends Component {
 | `checkedChange` | Emits if the user interacts with the button. If the button is disabled or checked, the event will not be emitted. | `CustomEvent<any>` |
 
 
-## CSS Custom Properties
```
janivo, line 70: This probably did not happen on purpose.

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
```diff
@@ -29,6 +30,12 @@ export class Snackbar implements ComponentInterface {
    */
   @Prop() inoAlignment?: 'left' | 'right' | 'center' = 'center';
 
+  /**
+   * Changes the snackbar type
+   * Valid options are: warning, error
```
janivo, line 35: The options don't need to be mentioned explicitly. They will be generated

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
```diff
@@ -29,6 +30,12 @@ export class Snackbar implements ComponentInterface {
...
+   */
+  @Prop() inoType?: SnackbarType;
```
janivo, line 37: If we use a literal type, we should provide a default value. Maybe add another option `default` which is the default value of `inoType` (analogous to line 31).

packages/elements/src/components/ino-snackbar/ino-snackbar.scss
```diff
@@ -58,3 +60,11 @@ ino-snackbar {
     }
   }
 }
+
+@each $type in $snackbar-types {
+  ino-snackbar[ino-type="#{$type}"] {
```
janivo, line 65: We should provide CSS vars to customize the color of these different types

janivo (changes requested):

packages/elements/src/components/ino-snackbar/ino-snackbar.scss
P1NHE4D: ✅

packages/elements/src/components/ino-segment-button/readme.md
P1NHE4D: Nope - this happened by accident. Should be fixed now.

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
P1NHE4D: ✅

janivo (approved):

PR #120, opened 2020-10-07
feat(elements|ino-card): add property to disable hover animation

### LINKED ISSUE #65: ino-card: Provide a way to disable the hover effect
When using the `ino-card` there is currently no way to disable the hover effect. In some cases (e.g. just displaying some data), it might be helpful to disable this behavior.

Provide a way to disable the hover effect which elevates the card.

I've tried to disable/overwrite the current behaviour with CSS but without any success.

P1NHE4D:
✨ **Features:**
* Added a new `inoDisableHover` property to the `ino-card` component that can be used to disable the hover effect

**Closes #65**

### DIFF
--- a/packages/elements/src/components/ino-card/ino-card.tsx
+++ b/packages/elements/src/components/ino-card/ino-card.tsx
@@ -18,7 +18,7 @@ export class Card implements ComponentInterface {
   /**
    * Disables the hover effect if true
    */
-  @Prop() inoDisableHover: boolean = false;
+  @Prop() inoDisableElevation: boolean = false;
 
   render() {
     const classList = classnames({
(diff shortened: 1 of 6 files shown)


packages/elements/src/components/ino-card/ino-card.tsx
```diff
@@ -15,6 +15,11 @@ export class Card implements ComponentInterface {
    */
   @Prop() inoSelected: boolean = false;
 
+  /**
+   * Disables the hover effect if true
+   */
+  @Prop() inoDisableHover: boolean = false;
```
janivo, line 21: Maybe we should call it `inoDisableElevation`? That's how it's called in [Material-UI](https://material-ui.com/components/buttons/#contained-buttons)

P1NHE4D: That's better. I'll rename it.

janivo (approved):

PR #123, opened 2020-10-07
fix(elements|ino-select): fix positioning while scrolling

### LINKED ISSUE #37: ino-select: opened menu doens't close on scroll
### Describe the bug
If you open the select box and scroll the site the menu is kept open.

### To Reproduce
1. Go to 'https://elements.inovex.de/dist/latest/storybook/?path=/story/input-ino-select--default-usage'
2. Click on any 'ino-select' sample to open the options
3. Scroll down and up

### Expected behavior
According to the behavior of the native select the menu should close on scrolling the page.
Probably its a better option to replace the underlying material select with the native one? As there are further complexities with ino-select, propaply its a better solution to switch to the native select and replace the underlying material select https://github.com/inovex/elements/issues/26?

### LINKED ISSUE #46: ino-input: autofocus does not work inside ino-card
### Describe the bug

Autofocus of ino-input does not work if wrapped by an ino-card element. 

### To Reproduce
1. Put ino-input with autofocus  enabled inside ino-card element
2. Reload Page
3. ino-input is not focused

### Expected behavior
Autofocus should work despite its parent element.

JCofman:
Closes #37

This PR fixes the positioning of the select dropdown menu during user scroll. @janivo I think that this should not break any positioning logic but could you please verify whether I have missed something in this context.

We could introduce the ability to remove the scroll event. While It might be a good Idea I also think this can also be implemented in the consuming framework. For example in React with https://github.com/theKashey/react-remove-scroll#readme


JCofman:
Does not introduce new error 
james-frontend

janivo:
@JCofman: Could you resolve the conflicting files?

JCofman:
done 👍

janivo (approved):

PR #156, opened 2020-10-27
fix(elements|ino-fab): ensure creation of tooltip after fab loaded

### LINKED ISSUE #155: ino-fab: Tooltip failing to append
### Describe the bug
If an `ino-tooltip` renders within `ino-fab`, it's not ensured that they load in a defined order. It comes down to a race-condition on which one loads first. That leads to various problems (anchor for tooltip not found, tooltip undefined, ...).

### To Reproduce
1. Use the `ino-fab` with tooltip enabled (not extended, tooltip-placement not `none`)
2. See console
3. Refresh if no error

### Possible Solution
Wait for the `ino-fab` to finish loading and append the `ino-tooltip` afterward.

janivo:
Closes #155 

## Proposed Changes

- Wait for `ino-fab` to finish loading
- Create and append `ino-tooltip` element afterward

### DIFF
--- a/packages/elements/src/components/ino-fab-set/readme.md
+++ b/packages/elements/src/components/ino-fab-set/readme.md
@@ -127,8 +127,8 @@ The ino-fab-set has a controlled (unmanaged) attribute `inoOpenDial`. For this r
 graph TD;
   ino-fab-set --> ino-fab
   ino-fab-set --> ino-icon
-  ino-fab --> ino-icon
   ino-fab --> ino-tooltip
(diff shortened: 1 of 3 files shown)


pfecht:
Did you ensure that the documentation remains the same as before? I thought the dynamic rendering changed the mermaid dependency graph.

janivo:
You are right, I forgot to commit the mermaid graph changes.

pfecht (approved):

PR #170, opened 2020-11-04
fix(elements-angular): prevent events from firing twice

### LINKED ISSUE #164: elements-angular: Event Handlers are called twice for checkbox component
### Describe the bug
Value Accessors for ino-checkbox handle the same events twice. Hence, `checkedChange`-handler sin Angular are triggered twice. This leads to problems, if we want to implement toggle logic without explicitly check the emitted state.

This bug has a high potential to affect other components like radio, switch, ... as well (I did not test it tho).

### To Reproduce
Integrate the `@inovex.de/elements-angular` package on v2 and add the `InoElementsModule` as import in the `AppModule` of an Angular application (aká setup Angular for Elements).

Hence, define a component as:

```ts
import { Component } from "@angular/core";

@Component({
  selector: "app-root",
  templateUrl: "./app.component.html",
  styleUrls: ["./app.component.css"]
})
export class AppComponent {
  title = "CodeSandbox";

  events = [];

  checkedChangeHandler($event) {
    this.events.push("Event triggered");
  }
}
```

with the response template:

```html
<!--The content below is only a placeholder and can be replaced.-->
<div>
  <h1>
    Welcome to {{ title }}!
  </h1>
  This is a test sandbox:
  <ino-checkbox (checkedChange)="checkedChangeHandler($event)"></ino-checkbox>

  <ul>
    <li *ngFor="let event of events">{{event}}</li>
  </ul>
</div>
```

Finally, check and uncheck the checkbox. As you can see, the event is thrown twice.

=> See the following codesandbox: https://codesandbox.io/s/eloquent-allen-fppgd

### Expected behavior
EventHandlers are only triggered once.

### Additional context
My insights are:

The event (`checkedChange`) is only thrown once from `ino-checkbox`, but handled twice within the `BooleanControlValueAccessor` In the https://github.com/inovex/elements/blob/master/packages/elements-angular/elements/src/directives/control-value-accesors/value-accessor.directive.ts we have the following code:

```ts
  handleChangeEvent(value: any) {
    if (value !== this.lastValue) {
      this.lastValue = value;
      this.onChange(value);
      this.writeValue(value);
    }
  }
```

Since this code is called twice for every checkbox state change, this leads to the following behavior:
1. Call: Value is different from lastValue => writes the value to the native input
2. Call: Value equals last value => do nothing.

However, the `checkedChange`-Event is still handled by the defined `checkedChange`-handler.

### (No) Possible Solution
It is not possible, to simply  change the abvoe code to:
```ts
  handleChangeEvent(value: any) {
    if (value !== this.lastValue) {
      this.lastValue = value;
      this.onChange(value);
      this.writeValue(value);
    } 
   else {
       // ===>  event.stopImmediatePropagation()
    }
  }
```

to avoid other handlers from being called, since Angular does not support this (https://github.com/angular/angular/issues/9587). 
Further, this "solution" doesn't answer the question of why the event is handled twice in the first place.

One possible path of investigation may be to use the native `change` and to check the difference of our implementation to ionic https://github.com/ionic-team/ionic-framework/tree/master/angular/src/directives/control-value-accessors

JCofman:
Closes #164 

## Proposed Changes
As mentioned https://github.com/ionic-team/stencil-ds-output-targets/issues/81 there is an issue with the `@stencil/angular-output-target` script. This is a patch till the issue gets resolved.

### DIFF
--- a/packages/elements-angular/elements/src/directives/angular-component-lib/utils.ts
+++ b/packages/elements-angular/elements/src/directives/angular-component-lib/utils.ts
@@ -1,4 +1,4 @@
-import { fromEvent } from 'rxjs';
+import { EventEmitter } from '@angular/core';
 
 export const proxyInputs = (Cmp: any, inputs: string[]) => {
   const Prototype = Cmp.prototype;
@@ -25,7 +25,7 @@ export const proxyMethods = (Cmp: any, methods: string[]) => {
 };
 
 export const proxyOutputs = (instance: any, el: any, events: string[]) => {
-  events.forEach((eventName) => (instance[eventName] = fromEvent(el, eventName)));
+  events.forEach((eventName) => (instance[eventName] = new EventEmitter()));
 };
 
 // tslint:disable-next-line: only-arrow-functions
--- a/packages/elements-angular/elements/src/directives/proxies-utils.ts
+++ b/packages/elements-angular/elements/src/directives/proxies-utils.ts
@@ -1,46 +1,47 @@
 /* eslint-disable */
 /* tslint:disable */
-import { fromEvent } from 'rxjs';
+import { EventEmitter } from '@angular/core';
 
 export const proxyInputs = (Cmp: any, inputs: string[]) => {
   const Prototype = Cmp.prototype;
   inputs.forEach(item => {
-  Object.defineProperty(Prototype, item, {
-    get() {
-    return this.el[item];
-    },
-    set(val: any) {
-    this.z.runOutsideAngular(() => (this.el[item] = val));
-    }
-  });
+    Object.defineProperty(Prototype, item, {
+      get() {
+        return this.el[item];
+      },
+      set(val: any) {
+        this.z.runOutsideAngular(() => (this.el[item] = val));
+      }
+    });
   });
 };
 
 export const proxyMethods = (Cmp: any, methods: string[]) => {
   const Prototype = Cmp.prototype;
   methods.forEach(methodName => {
-  Prototype[methodName] = function () {
-    const args = arguments;
-    return this.z.runOutsideAngular(() =>
-    this.el[methodName].apply(this.el, args)
-    );
-  };
+    Prototype[methodName] = function() {
+      const args = arguments;
+      return this.z.runOutsideAngular(() =>
+        this.el[methodName].apply(this.el, args)
+      );
+    };
   });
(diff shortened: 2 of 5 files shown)


janivo (dismissed):

janivo:
@pfecht: Could you check if this fixes your problem?

pfecht:
@janivo  The problem still exists after the fix. My guess is that the changed `proxy-utils.ts`is not used at all but the `elements/packages/elements-angular/elements/src/directives/angular-component-lib.utils.ts` instead? 

Why do we have this two similar files?

JCofman:
@pfecht could you describe what kind of issues you had? 

I quickly created two demos 

*without the fix*
ohne-fix

*with the fix*
mit-fix

janivo:
Same as @pfecht. It seems like the `packages/elements-angular/elements/src/directives/angular-component-lib/utils.ts` is automatically rewritten from `import { EventEmitter } from '@angular/core';` to `import { fromEvent } from 'rxjs';` with every build and then distributed, so your changes don't have an effect.

janivo:
@pfecht recently improved the whole `yalc` process. Could you rebase this branch to master and try out your changes again @JCofman? If it's working for you, maybe you can provide the steps you took to make it work. 

My steps:
1. Rebased onto the `master` branch
2. Run `yarn publish:yalc` (selected `Angular`)
3. Run  `yarn link:elements` in `elements-example-angular`
4. Start the angular application

JCofman:
Hey @janivo and @pfecht 

I have rebased the branch and saw the same issue. I am currently not sure about which files are generated automatically by stencil and the whole build process. But as I saw that we have an auto-generated  `angular-component-lib/utils.ts` file which we need to overwrite as well. I have rebased and pushed the fix. It works fine now if I do the same steps as @janivo 


My steps:
1. Run `yarn publish:yalc` (selected `Angular`)
3. Run  `yarn link:elements` in `elements-example-angular`
4. Start the angular application

janivo (approved):
It's working for me! Can you check as well, @pfecht?

pfecht:
Sorry, for the late response, works for me like a charm now. Thanks guys!

janivo:
I duplicated the branch in #201 to run checks. This did not work because Jacob worked on a fork. Thus I'm closing this PR and will merge #201.

PR #175, opened 2020-11-07
feat(elements|ino-input): add readonly textfield and datepicker

pfecht:
Adds a readonly state to the input field and datepicker.

Requires some attention! Colors behavior changed:
* Floating Label Color changed
* Select has different label color than

### DIFF
--- a/packages/elements/src/components/ino-label/ino-label.scss
+++ b/packages/elements/src/components/ino-label/ino-label.scss
@@ -1,7 +1,7 @@
-@use '../styles/mdc-customize';
+@use 'styles/mdc-customize';
+@use 'styles/colors';
 @use '@material/floating-label/mdc-floating-label';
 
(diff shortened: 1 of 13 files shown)


janivo:
@pfecht: Are you still working on this one?

packages/elements/src/components/ino-label/ino-label.scss
```diff
@@ -1,7 +1,7 @@
-@use '../styles/mdc-customize';
+@use 'styles/mdc-customize';
+@use 'styles/colors';
```
silentHoo, line 2: As we don't use colors inside this file, we don't need to import the colors file.

silentHoo (commented):
I think it's not good to float the label when the input is read only and there's no value set. Can we only float when it's editable?

https://user-images.githubusercontent.com/1610894/111140903-29987780-8583-11eb-8089-cac6209989ed.mp4

PR #189, opened 2020-11-17
feat(elements|ino-chip): add slot for icons and mark ino-icon prop as deprecated

### LINKED ISSUE #188: ino-chip: Use slots instead of `ino-icon` property
If you want to use an `ino-icon` inside of an `ino-chip`, you have to pass the icon name per the `ino-icon` property. This limits the flexibility of the icon because you cannot use all props of the `ino-icon` component (e.g. `src`). We should provide a slot for the icon, so we can use the whole API of the `ino-icon` component.

This solution is already implemented e.g. in the `ino-button`.

P1NHE4D:
✨ **Features:**
* Removed the `ino-icon` property
* Added a slot with the name `ino-icon-leading
* Adjusted the `ino-icon` and `ino-chip` stories accordingly

🎟️ **Closes #188**


janivo:
Maybe we should support both variants for now and print a warning that this feature (using the `ino-icon` prop) will be deprecated in the future. Otherwise, we might have to release a new major version again in such a short period of time. What do you think?

P1NHE4D:
> Maybe we should support both variants for now and print a warning that this feature (using the `ino-icon` prop) will be deprecated in the future. Otherwise, we might have to release a new major version again in such a short period of time. What do you think?

This is a great idea. I'll add it right away.

JCofman (approved):
LGTM :)

packages/elements/src/components/ino-chip/ino-chip.tsx
```diff
@@ -26,7 +26,13 @@ export class Chip implements ComponentInterface {
   @Prop() inoFill?: ChipSurface = 'solid';
 
   /**
-   * If enabled, prepends the slotted icon to the chip
+   * Prepends an icon to the chip label.
+   * @deprecated This property is deprecated and will be removed with the next major release. Instead, use the ino-icon-leading slot.
```
janivo, line 30: I would suggest that we also check if the `inoIcon` property is used initially (e.g. in `componentDidMount`) and print a warning with `console.warn` with the deprecated message above. This way we make sure that everyone is informed even if he's not checking the documentation regularly.

janivo (changes requested):

packages/elements/src/components/ino-chip/ino-chip.tsx
P1NHE4D: Good idea

janivo (approved):

PR #198, opened 2020-11-19
fix(elements|ino-input): add autofocus prop to input element

### LINKED ISSUE #46: ino-input: autofocus does not work inside ino-card

JCofman:
Closes #46 

Currently, we do not pass down the autofocus prop to the input element.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "husky": "^1.3.1",
     "js-yaml": "^3.14.0",
     "lint-staged": "^8.1.5",
-    "prettier": "^1.16.4",
+    "prettier": "^2.1.2",
     "prompts": "^2.4.0",
     "shelljs": "^0.8.4",
     "shx": "^0.3.2",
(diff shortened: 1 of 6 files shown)


janivo (commented):
Were there any issues for you with the autofocus functionality? Currently, we set the autofocus in line 269 like this:
```tsx
if (this.autoFocus) {
  this.textfield.focus();
}
```

janivo:
Could you also resolve the conflicts?

janivo:
Did you check if this fix resolves the problems mentioned in the linked issue?

JCofman:
As far as I have understood the problem this should solve the issue. I have also tested it locally ;)

janivo (approved):

PR #199, opened 2020-11-19
feat(elements|ino-input-file): add drag and drop window

### LINKED ISSUE #76: ino-input-file: Drag 'n Drop
I would like to upload a file by dragging it onto the `ino-input-file` component.

I can already upload a file by clicking the upload button and selecting the file within the file explorer. While this solution is working correctly, it's sometimes more comfortable to drag the file without having to navigate in the file explorer.

P1NHE4D:
✨ **Features:**
* Added the ability to upload files using a drag and drop window to the `ino-input-file` component

ℹ️ **FYI:**
* There's currently a bug that will result in the event emitter used by the `ino-input-file` component to emit multiple events at once upon setting the action knobs in Storybook to a different value. This bug is also present in other components, such as the `ino-snackbar`(see #81). 

🎟️ **Closes #76**

### DIFF
--- a/packages/elements/src/components/ino-input-file/ino-input-file.tsx
+++ b/packages/elements/src/components/ino-input-file/ino-input-file.tsx
@@ -67,7 +67,6 @@ export class InputFile implements ComponentInterface {
   }>;
 
   componentDidLoad(): void {
-    // TODO: check if browser supports dragAndDrop
     this.configureDragAndDrop();
   }
 
--- a/packages/elements/src/components/ino-input-file/ino-input-file.tsx
+++ b/packages/elements/src/components/ino-input-file/ino-input-file.tsx
@@ -8,6 +8,8 @@ import classNames from 'classnames';
 export class InputFile implements ComponentInterface {
   @Element() el!: HTMLElement;
 
+  private eventListeners: [string, EventListener | EventListenerObject][] = [];
+
   /**
    * The types of files accepted by the server.
    */
@@ -63,11 +65,26 @@ export class InputFile implements ComponentInterface {
    */
   @Event() changeFile!: EventEmitter<{
     e: any;
-    files: object[];
+    files: File[];
   }>;
 
   componentDidLoad(): void {
     this.configureDragAndDrop();
+  }
+
+  disconnectedCallback(): void {
+    this.eventListeners.forEach((tuple) => this.removeEventListeners(this.el, tuple[0], tuple[1]));
+  }
+
+  private addEventListeners(el: HTMLElement, events: string, fn: EventListener | EventListenerObject): void {
+    this.eventListeners.push([events, fn]);
+    events.split(' ').forEach(e => {
+      el.addEventListener(e, fn);
+    });
+  }
+
+  private browserSupportsDragAndDrop(): boolean {
+    return (('draggable' in this.el) || ('ondragstart' in this.el && 'ondrop' in this.el) && 'FormData' in window && 'FileReader' in window);
   }
 
   private configureDragAndDrop(): void {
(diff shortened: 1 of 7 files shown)


packages/elements/src/components/ino-input-file/ino-input-file.tsx
```diff
@@ -50,6 +66,57 @@ export class InputFile implements ComponentInterface {
     files: object[];
   }>;
 
+  componentDidLoad(): void {
+    this.configureDragAndDrop();
+  }
+
```
JCofman, line 72: I might be wrong ;) but should we also remove the initialized event listeners on `componentDidUnload`?

packages/elements/src/components/ino-input-file/ino-input-file.tsx
```diff
@@ -50,6 +66,57 @@ export class InputFile implements ComponentInterface {
...
+        this.el,
+        'dragend dragleave drop',
+        () => box.classList.remove('ino-input-file__dnd-dragover')
+      );
+      this.el.addEventListener('drop', (e) => {
+        if (this.disabled) {
+          return;
+        }
+        console.log('Drop');
+        const files = e.dataTransfer.files as FileList;
```
JCofman, line 102: I think we could remove the as FileList type conversion

packages/elements/src/components/ino-input-file/ino-input-file.tsx
```diff
@@ -50,6 +66,57 @@ export class InputFile implements ComponentInterface {
     files: object[];
```
JCofman, line 66: Instead of the object Type I think we could type it `File[]`

packages/elements/src/components/ino-input-file/ino-input-file.tsx
```diff
@@ -50,6 +66,57 @@ export class InputFile implements ComponentInterface {
...
+        () => {
+          box.classList.add('ino-input-file__dnd-dragover');
+        }
+      );
+      this.addEventListeners(
+        this.el,
+        'dragend dragleave drop',
+        () => box.classList.remove('ino-input-file__dnd-dragover')
+      );
+      this.el.addEventListener('drop', (e) => {
```
JCofman, line 97: I would suggest to explicitly type this `e: DragEvent` :)

JCofman (changes requested):

packages/elements/src/components/ino-input-file/ino-input-file.tsx
```diff
@@ -50,6 +66,57 @@ export class InputFile implements ComponentInterface {
...
+        if (this.disabled) {
+          return;
+        }
+        console.log('Drop');
```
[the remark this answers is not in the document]
P1NHE4D: Whoops 😂

packages/elements/src/components/ino-input-file/ino-input-file.tsx
P1NHE4D: Yes, I think it's better we do so.

packages/elements/src/components/ino-input-file/ino-input-file.tsx
P1NHE4D: ✅

packages/elements/src/components/ino-input-file/ino-input-file.tsx
P1NHE4D: ✅

packages/elements/src/components/ino-input-file/ino-input-file.tsx
P1NHE4D: ✅

JCofman (approved):

Issue #202, opened 2020-11-23
ino-nav-drawer: Background-Color changed unexpectedly


Sl1nd:
### Describe the bug
The Background-Color of `ino-nav-drawer` changed in v2

### To Reproduce
compare 
* [v2.0.0](https://elements.inovex.de/dist/v2.0.0/storybook/?path=/story/structure-ino-nav-drawer--default-usage ) 
* [v1.2.0](https://elements.inovex.de/dist/v1.2.0/storybook/?path=/story/structure-ino-nav-drawer--default-usage)

PR #203, opened 2020-11-23
style(elements|ino-nav-item): restore original hover and active behavior

Sl1nd:
Closes #
https://github.com/inovex/elements/issues/202

## Proposed Changes

- Remove Ripple effect

### DIFF
--- a/packages/elements/src/components/ino-list-item/ino-list-item.tsx
+++ b/packages/elements/src/components/ino-list-item/ino-list-item.tsx
@@ -81,6 +81,7 @@ export class ListItem implements ComponentInterface {
     return (
       <Host>
         <li class={listItemClasses}>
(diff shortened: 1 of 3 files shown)


janivo (changes requested):
These changes disable the hover functionality of all `ino-list-item`s. I'll check for an alternative

janivo:
@Sl1nd could you check if this works for you?

Sl1nd:
I realized that also the `ino-list`checkbox behavior unintentionally changed in relation to the ripple Effect. Should it be fixed in that context as well?

janivo:
No, I think this should be part of another PR

janivo (approved):

janivo (changes requested):
The CSS Props have been removed for some reason

janivo (approved):

PR #209, opened 2020-11-24
refactor(elements|ino-input): cleanup styles

janivo:
Cleaned up and restructured the `ino-input.scss` in the hope to make it easier to work on. I structured the stylesheet by text-field variants (default, filled, outlined). There also were some new mixins with the new mdc versions which replaced some complicated selectors.

### DIFF
--- a/packages/elements/src/components/ino-input/ino-input.scss
+++ b/packages/elements/src/components/ino-input/ino-input.scss
@@ -42,7 +42,7 @@ $icon-padding-to-border: 6px;
   }
 
   /* Firefox */
-  input[type=number] {
+  input[type='number'] {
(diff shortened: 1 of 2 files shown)


janivo:
This PR also fixes a problem with the positioning of icons

silentHoo:
Code looks good, but there are some missing styles for the disabled state. I've attached the current disabled Styling from `v2.3.0`. And the cursor on the clickable icon is not set to `cursor: pointer`.

image

janivo:
Thanks for the review @silentHoo!
I rebased onto the current master and applied your feedback. Could you check again?

silentHoo (approved):

PR #221, opened 2020-12-02
feat(elements|ino-chip): add slot for trailing icon

### LINKED ISSUE #204: ino-chip: allow custom trailing icons
Currently, it is only possible to position a custom icon at the leading position of an `ino-chip`. We should provide more flexibility so the consumer is able to position the icon at the trailing position as well.

DoD's:
- Custom icons can be placed at the trailing position (default should be leading)
- The trailing icon does not collide with the `inoRemovable` option (maybe the removable option will be ignored if a trailing icon is provided?)

P1NHE4D:
✨ **Features:**
* Added `ino-icon-leading` slot to the `ino-chip` component

🎟️ **Closes #204**

### DIFF
--- a/packages/elements/src/components/ino-chip/ino-chip.scss
+++ b/packages/elements/src/components/ino-chip/ino-chip.scss
@@ -15,6 +15,18 @@ ino-chip {
   .mdc-chip {
     @include chips.fill-color-accessible($default-chip-color);
 
+    .mdc-chip__icon--trailing {
+      width: 20px;
+      height: 20px;
+      font-size: 20px;
+    }
+
+    ino-icon {
+      --icon-width: .75em;
+      --icon-height: .75em;
+      --icon-color: currentColor;
+    }
+
     ino-icon {
       --icon-width: 0.75em;
       --icon-height: 0.75em;
--- a/packages/elements/src/components/ino-chip/ino-chip.scss
+++ b/packages/elements/src/components/ino-chip/ino-chip.scss
@@ -22,12 +22,6 @@ ino-chip {
     }
 
     ino-icon {
-      --icon-width: .75em;
-      --icon-height: .75em;
-      --icon-color: currentColor;
-    }
-
-    ino-icon {
       --icon-width: 0.75em;
       --icon-height: 0.75em;
       --icon-color: currentColor;
--- a/packages/elements/src/components/ino-chip/ino-chip.tsx
+++ b/packages/elements/src/components/ino-chip/ino-chip.tsx
@@ -147,7 +147,7 @@ export class Chip implements ComponentInterface {
           <span class="mdc-chip__text">{this.inoLabel}</span>
 
           {this.inoIconTrailing && (
-            <span class="mdc-chip__icon mdc-chip__icon--trailing">
+            <span class="mdc-chip__icon ">
               <slot name="ino-icon-trailing"/>
             </span>
           )}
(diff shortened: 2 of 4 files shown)


packages/elements/src/components/ino-chip/ino-chip.scss
```diff
@@ -15,6 +15,18 @@ ino-chip {
   .mdc-chip {
     @include chips.fill-color-accessible($default-chip-color);
 
+    .mdc-chip__icon--trailing {
+      width: 20px;
+      height: 20px;
+      font-size: 20px;
+    }
+
+    ino-icon {
```
janivo, line 24: This seems to be a duplication of lines 24 - 28, no?

packages/elements/src/components/ino-chip/ino-chip.scss
```diff
@@ -15,6 +15,18 @@ ino-chip {
   .mdc-chip {
     @include chips.fill-color-accessible($default-chip-color);
 
+    .mdc-chip__icon--trailing {
```
janivo, line 18: Why is this necessary? Do the leading and trailing icons not have the same size?

packages/elements/src/components/ino-chip/ino-chip.tsx
```diff
@@ -36,6 +36,11 @@ export class Chip implements ComponentInterface {
    */
   @Prop() inoIconLeading = false;
 
+  /**
+   * If enabled, appends the slotted icon to the chip label
```
janivo, line 40: We also should mention that this flag will disable the `ino-removable` functionality

janivo (changes requested):

janivo:
Hey @P1NHE4D, could you have a look at this again?

P1NHE4D:
> Hey @P1NHE4D, could you have a look at this again?

I'm on it right now. 🙂

packages/elements/src/components/ino-chip/ino-chip.scss
P1NHE4D: apparently 😄

packages/elements/src/components/ino-chip/ino-chip.tsx
P1NHE4D: ✅

packages/elements/src/components/ino-chip/ino-chip.scss
P1NHE4D: Removing the class results in trailing icons being smaller than leading icons. We're also using the 'mdc-chip__icon--leading' class for the leading icon. However, due to the absence of a corresponding class for trailing icons in the mdc package, I added the 'mdc-chip__icon--trailing' class manually. I also attached a screenshot to illustrate the effect of removing the aforementioned.
Skjermbilde 2021-02-03 kl  10 50 49

janivo: Alright, thanks for the explanation ✔️

janivo (changes requested):
Overall this seems to collide with the changes made in https://github.com/inovex/elements/pull/277 as we removed all `inoIconLeading/Trailing` slots and check for them at runtime. 
Could you rebase on the #277 and update the render function accordingly?

janivo:
Ah, my bad here. I thought something went wrong while rebasing on master, so I tried it myself and force-pushed. I realized that I told you to rebase on the #277 and that it's not merged yet 🤦🏼‍♂️

janivo:
I'll review as soon as #277 is merged

janivo:
@P1NHE4D could you rebase on last time? 😅

janivo (approved):
looking good!

PR #223, opened 2020-12-03
feat(elements|ino-popover): add interactive property

janivo:
## Proposed Changes

- Adds a new property `inoInteractive` which is disabled by default
- The new prop enables interaction inside the popover (e.g. clicking a button)
- This did not work prior because events wouldn't be triggered


AlessaRad (approved):
Nice work!

PR #226, opened 2020-12-07
fix(elements|ino-datepicker): use date format to improve validation

### LINKED ISSUE #225: InoDatepicker: "Cannot read property 'parseDate' of undefined"
### Describe the bug
InoDatepicker is set to invalid although the value has the proper format. 

### To Reproduce
Difficult to reproduce. Used `InoDatepicker` with React. The Error occured irregular on rendering the `InoDatepicker` within a Modal.

### Expected behavior
The invalid state should only be set if format is not correct.

### Your Environment
 

| Software | Version |
| -------- | ------- |
| @inovex.de/elements | v2.2.0
| Browser | Chrome
| Operating System | MacOS

### Possible Solution
It seems the flatpicker instance isnt't instantiated at the time of validating the value.

Sl1nd:
Closes #225 

- A debounce was not possible, because of the multiple possibilities to enter values (e.g. copy/paste, flatpicker)
- It also turned out that validating on blur leads to a worse user experience
- Only validate if flatpicker is set
- Also ino-range validation is fixed

### DIFF
--- a/packages/elements/src/components/ino-datepicker/ino-datepicker.e2e.ts
+++ b/packages/elements/src/components/ino-datepicker/ino-datepicker.e2e.ts
@@ -158,12 +158,12 @@ describe('InoDatepicker', () => {
       const inoDatepickerEl = await page.find(DATEPICKER);
       const buttonEl = await page.find('button');
       const flatpickrInputEl = await page.find('.flatpickr-input');
-    
+
       inoDatepickerEl.setAttribute('required', true);
       await page.waitForChanges();
 
       expect(flatpickrInputEl).not.toHaveClass('mdc-text-field--invalid');
-      
+
       inoDatepickerEl.click();
       await page.waitForChanges();
       buttonEl.click();
@@ -181,10 +181,10 @@ describe('InoDatepicker', () => {
 
       await input.type('1');
       await page.waitForChanges();
-      
+
       expect(valueChangeEvent).toHaveReceivedEvent();
     });
-    
+
     it('should not emit a valueChange event if datepicker is disabled', async () => {
       let page = await setupPageWithContent(INO_DATEPICKER);
       const datepicker = await page.find(DATEPICKER);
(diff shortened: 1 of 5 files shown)


packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -336,17 +341,30 @@ export class Datepicker implements ComponentInterface {
...
+    };
+  }
+
+  private validateInput(value: string): boolean {
+    let formattedDate: string;
+    let parsedDate: Date; 
+    parsedDate = this.flatpickr.parseDate(value);
+    formattedDate = this.flatpickr.formatDate(parsedDate, this.flatpickr.config.dateFormat);
     
+    return formattedDate !== value ?  false : true;
```
janivo, line 367: This can be simplified to `return formattedDate === value`, no?

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -336,17 +341,30 @@ export class Datepicker implements ComponentInterface {
...
     }
+
+    if(!value && !this.required) {
+      this.isValid = true; 
+      return;
+    };
+  }
+
+  private validateInput(value: string): boolean {
+    let formattedDate: string;
```
janivo, line 362: We should use `const` if we do not have to use `let`

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -336,17 +341,30 @@ export class Datepicker implements ComponentInterface {
   });
 
   private setValidState(value: string): void {
-    let formattedDate: string;
-    let parsedDate: Date;
+    if(this.inoRange) {
+      let arr = value.match(/(\d[^a-z]+)/g).map(match => this.validateInput(match.trim()));
```
janivo, line 345: We should use `const` if we do not have to use `let`

janivo, line 345: Could you extract that regex as a static member and give it a name? This way, it is more clear what the regex is supposed to test.

janivo, line 345: The name `arr` is kinda generic. Could you name it more spelling?

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -58,9 +58,14 @@ export class Datepicker implements ComponentInterface {
 
   @Watch('value')
   valueChanged(value: string) {
-    this.setValidState(value);
-    if (this.flatpickr) {
-      this.flatpickr.setDate(value, false, this.inoDateFormat);
+    try {
+      if (this.flatpickr) {
+        this.setValidState(value);
+        this.flatpickr.setDate(value, false, this.inoDateFormat)
```
janivo, line 64: With this logic, the flatpickr will always set the date, even if it is invalid. Is this intended?

packages/elements/src/components/ino-datepicker/ino-datepicker.e2e.ts
```diff
@@ -152,19 +153,22 @@ describe('InoDatepicker', () => {
       expect(datepicker).toHaveAttribute('ino-show-label-hint');
     });
 
-    it('should set invalid state for required inputs', async () => {
-      const page = await setupPageWithContent(INO_DATEPICKER);
+    it.only('should set invalid state for required inputs', async () => {
```
janivo, line 156: Why the `only`?

janivo (changes requested):

janivo (approved):

PR #237, opened 2020-12-23
feat(elements|ino-textarea): add filled style option

### LINKED ISSUE #142: ino-textarea: Support filled textarea styles
The mdc-textfield now supports outlined and ripple effects for textareas. As for now, we inovex elements only support the outlined textarea.

Would be great to have the same opportunities just like a regular ino-input 

see https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield#textarea

AlessaRad:
Closes #142

## Proposed Changes

- Adds the filled text field style option for the ino-textarea (analogous to ino-input)
- The outlined text field style temporarily remains the default style (to give consumers a heads up), but can be changed to filled by setting the inoOutlined property to false


packages/elements/src/components/ino-textarea/ino-textarea.tsx
```diff
@@ -139,6 +144,8 @@ export class Textarea implements ComponentInterface {
     if (this.autoFocus) {
       this.textfield.focus();
     }
+
+    console.warn(`The ino-textarea default type will be changed to 'filled' in the next major release (analogous to the ino-input). In order to use the 'outline' style, set the new 'inoOutline' property explicitly to true, please.`);
```
janivo, line 148: IMO we should only print this warning if the new `inoOutline` property is not used.

packages/elements/src/components/ino-label/ino-label.scss
```diff
@@ -23,14 +23,6 @@ ino-label {
     }
   }
 
-  &:not([ino-outline]) .mdc-floating-label {
```
janivo, line 26: These changes break the filled `ino-select` styles. If there are specific styles that should be applied to the label of the filled textarea, we should overwrite them in the textarea stylesheet.

janivo: image

janivo (changes requested):

packages/elements/src/components/ino-textarea/ino-textarea.tsx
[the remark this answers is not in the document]
AlessaRad: Done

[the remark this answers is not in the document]
AlessaRad: Done

AlessaRad: Done

packages/elements/src/components/ino-label/ino-label.scss
AlessaRad: Ok, I adjusted this as you suggested.

janivo (approved):

PR #242, opened 2021-01-05
fix(elements): replace deprecated componentWillUnload lifecycle method

### LINKED ISSUE #234: componentWillUnload is not triggered
### Describe the bug
The components use the `componentWillUnload` method for cleaning up. However, it appears that the method has been removed (?) from Stencil and thus not triggered at all. Instead, the `disconnectedCallback` should be used and works just fine.

### To Reproduce
For example on the datepicker
1. Add a datepicker
2. Remove the datepicker ==> The flatpickr is still available in the DOM
3. Readd the datpicker => A new flatpickr is added to the DOM
4. ...

### More details
As I just saw, the datepicker does not clean up at all. However, this does not change the problem itself.

Add 

```

  componentWillUnload() {
    this.dispose();
  }
```

==> Still the same behaviour

However, add 

```
  disconnectedCallback() {
    this.dispose();
  }
```

works just fine.

AlessaRad:
Closes #234 

## Proposed Changes

- substitute all deprecated componentWillUnload() calls with the disconnectedCallback() lifecycle method (as proposed in the bug description)
- remove flatpickr from DOM when ino-datepicker component is disconnected

The latter can be tested as follows:
- add this code snipped at the end of the ino-datepicker.stories.js file
 `<button id="removeDatepicker" onclick="console.log('remove all datepickers'); var datepickers = document.querySelectorAll('ino-datepicker'); Array.from(datepickers).forEach(function (datepicker) { datepicker.remove(); });">Remove all datepickers</button>`
- when clicking the button in the ino-datepicker story within storybook, all appearances of the flatpickr in the DOM tree (shown in the screenshot) should be removed
Screenshot 2021-01-07 at 11 29 49


janivo (changes requested):
We should always make sure that the element is defined before we destroy them. Otherwise, we might run into runtime errors. Could you implement such a condition if not already exists?

We could create a helper function like this and use it in every component where we call the destroy function:
```ts
destroyIfExists(el?: any): void { // instead of any, is there a generic mdc element type?
	el && el.destroy();
}
```

The helper function might be overkill, not quite sure

AlessaRad:
> We should always make sure that the element is defined before we destroy them. Otherwise, we might run into runtime errors. Could you implement such a condition if not already exists?
> 
> We could create a helper function like this and use it in every component where we call the destroy function:
> 
> ```ts
> destroyIfExists(el?: any): void { // instead of any, is there a generic mdc element type?
> 	el && el.destroy();
> }
> ```
> 
> The helper function might be overkill, not quite sure

Good idea! I inserted all the corresponding checks. I didn't use a helper function though, because it is quite short (and probably more performant) with the optional chaining operator (e.g. this snackbarInstance?.destroy()).

packages/elements/src/components/ino-menu/ino-menu.tsx
```diff
@@ -55,7 +55,7 @@ export class Menu implements ComponentInterface {
 
   disconnectedCallback() {
     if (this.menu) {
```
janivo, line 57: This check can be removed

janivo (changes requested):
Good idea! I forgot about the ? opterator

packages/elements/src/components/ino-menu/ino-menu.tsx
AlessaRad: Indeed!

packages/elements/src/components/ino-tooltip/ino-tooltip.tsx
```diff
@@ -80,10 +80,6 @@ export class Tooltip implements ComponentInterface {
     await this.create();
   }
 
-  async disconnectedCallback() {
-    await this.dispose();
```
janivo, line 84: Why did you remove this call?

AlessaRad: This call would destroy the tooltip right after the first hover/click (when it is removed from the DOM), so the tooltip would only be visible once. The tooltip should only be destroyed when the target element is destroyed (this is not covered in this MR). Same for the popover.

janivo (approved):
Good work!

PR #245, opened 2021-01-07
fix(elements|ino-datepicker): prevent interaction with disabled datepicker and fix validation

### LINKED ISSUE #205: InoInput and Datepicker change their validation state on disabled
### Describe the bug
The `ino-input` changes it input validation states provided by the `ino-input`, regardless of whether the input element is disabled or not. This leads to strange a behaviour for the datepicker which marks a disabled input field having a provided value as invalid.

Further, the datepicker icon should not be clickable if disabled.

### To Reproduce
1. Use a datpicker, provide a value, check the validation state.

### Expected behavior
One can not influence the input validation state on disabled inputs.

### Possible Solution
* Abort handling inoError if the element is disabled in https://github.com/inovex/elements/blob/master/packages/elements/src/components/ino-input/ino-input.tsx#L217
* Move the logic of https://github.com/inovex/elements/blob/master/packages/elements/src/components/ino-input/ino-input.tsx#L278 to the handler, and call the handler on init as well. This is oversimplified right now.

AlessaRad:
Closes #205 

## Proposed Changes

- abort click and error handling when datepicker is disabled (as suggested in the issue description)

### DIFF
--- a/packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
+++ b/packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
@@ -58,6 +58,7 @@ export class Datepicker implements ComponentInterface {
 
   @Watch('value')
   valueChanged(value: string) {
+    if (this.disabled) return;
 
     try {
       if (this.flatpickr) {
@@ -67,7 +68,7 @@ export class Datepicker implements ComponentInterface {
       if (this.flatpickr && this.isValid) {
         this.flatpickr.setDate(value, false, this.inoDateFormat)
       }
-    } catch(e) {
+    } catch (e) {
       // Input could not be parsed e.g. empty spaces
       this.isValid = false;
     }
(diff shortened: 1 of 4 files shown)


packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -129,12 +130,17 @@ export class Datepicker implements ComponentInterface {
     this.create();
   }
 
+  @Watch('disabled')
+  disabledChanged() {
+    this.dispose();
+  }
```
janivo, line 136: There are two use-cases that need to be addressed here:
- Datepicker will be disabled -> Datepicker will be disposed
- Datepicker will be enabled -> Datepicker will be created

```suggestion
  @Watch('disabled')
  disabledChanged(newValue: boolean) {
  	if(newValue) {
  		this.dispose();
  	} else {
  		this.create();
    }
  }
```

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -293,6 +299,8 @@ export class Datepicker implements ComponentInterface {
   private static NUMBERS_WITH_SPECIAL_CHARS = /(\d[^a-z]+)/g
 
   private create() {
+    if (this.disabled) return;
```
janivo, line 302: This function should not be called if the disabled flag is set. We should move this to the `componentDidLoad` function.

janivo (changes requested):

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
AlessaRad: I agree, thanks for pointing it out!

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
AlessaRad: Done

janivo (approved):

PR #258, opened 2021-01-14
fix(elements|ino-progress-bar): restore buffer dots

### LINKED ISSUE #177: ino-progress-bar: Buffering dots not working
### Describe the bug
The buffering dots of the `ino-progess-bar` are not displayed anymore. This bug was introduced with version 2.0.0.

### To Reproduce
1. Start the Storybook
2. Go to Notification/ino-progess-bar
3. See variants without buffering dots

### Screenshots
Version 1.2.0:
image

Latest:
image

### Possible Solution
This was probably introduced with the MDC update to version 7.0.0. Maybe it's just a class that's missing.

AlessaRad:
Closes #177 

## Proposed Changes

- Fix progress bar styling and make use of provided mdc sass mixin 
- The styling of the buffer bar and dots should be configurable via the `buffer-color` mixin (`@include linear-progress.buffer-color((var(--progress-bar--buffer-color)));`), but it seems like this only styles the buffer bar correctly, but not the dots.

### DIFF
--- a/packages/elements/src/components/ino-progress-bar/ino-progress-bar.scss
+++ b/packages/elements/src/components/ino-progress-bar/ino-progress-bar.scss
@@ -10,17 +10,7 @@
    */
   --progress-bar--bar-color: #{colors.ino-color(primary)};
   --progress-bar--buffer-color: #{colors.ino-color(light, light)};
-}
 
-:host {
+  @include linear-progress.core-styles;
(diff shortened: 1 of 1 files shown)


janivo (approved):

janivo:
I found some problems which we should address in another issue: #264

PR #267, opened 2021-01-19
refactor(*): refactor sass

### LINKED ISSUE #159: Rewrite and clean-up Sass Files
At the moment it looks like there are some inconsistencies in the sass definition files.

### Possible Solution

- [x] Use Modul (`@use`) everywhere.

- [x] Delete `styles/variables.scss` since these are currently only z-index (of which 2 are not used). Also remove it from the `stencil.config.ts` since it does not make sense to make these as global styles.

- [x] Rethink the structure of the entire `styles` folder in terms of sass modules. Wouldn't it be better to have sth like `@use core/fonts` as imports

- [x] Make a decision either for relative path imports or not.

- [ ] Double-check if the mdc integration is still up-to-date (it seems that in recent versions the imports have changed).

pfecht:
Closes #159 

Refactors the entire sass structures to a more "sass modules"-native way.

## Proposed Changes
* Rename src/styles to src/base
* Move component specific sass (like tooltip.scss) to the component
* Use absolute instead of relative imports
* Rename styles/colors to base/theme and styles/font to base/typography.
* Refactor mixins/functions to remove the obsolete `ino`-prefix
* Rename styles/mixins to base/animation and rename the sass mixins for naming conventions.
* Remove unused variables.scss from stencil.ts


packages/elements/src/components/base/_typography.scss
```diff
@@ -0,0 +1,73 @@
+// ##################################
+// Font and typography
+// ##################################
+
+// => Font Family
+
+$_font-families: (
+  sans-serif: unquote("'Gilroy', Verdana, sans-serif"),
```
janivo, line 8: Where did you get these font-families from?
Did you c&p these by mistake?
We used `'Lato', Helvetica, Verdana, sans-serif` before.

janivo (commented):
pretty nice work overall!

packages/elements/src/components/base/_typography.scss
pfecht: True that. Fixed it, and renamed the mixins of spinner to the private`_` syntax (thanks @AlessaRad)

janivo (approved):

PR #269, opened 2021-01-22
chore: upgrade vulnerable highlight.js package from 9.18.1 to 10.5.0

janivo:
Upgraded highlight.js to a non-vulnerable version, see https://github.com/inovex/elements/security/dependabot


pfecht (approved):
👍  But why did you remove the `yarn.lock` of elements?

janivo:
@pfecht: I thought there is no reason to keep it because we only use the lock file on the root level, no?

pfecht:
Yeah, I think so too, if you made a `yarn` in the root and nothing changed it should be fine.

PR #271, opened 2021-01-25
fix(elements|ino-datepicker): fix random year skipping in month picker

### LINKED ISSUE #215: ino-monthpicker: Skipping next and previous years on arrow click
### Describe the bug

When navigating through years by clicking the arrow icons, a click will either jump over multiple previous or next years. Sometimes another month is marked as well.

### To Reproduce

buggyMonthpicker(1)

janivo:
Closes #215 

## Proposed Changes

- Remove the custom logic to handle arrow clicks
- Add tests

### DIFF
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -82,7 +82,7 @@
     "autosize": "^4.0.2",
     "classnames": "^2.2.6",
     "currency.js": "^1.2.2",
-    "flatpickr": "^4.4.6",
+    "flatpickr": "^4.6.9",
     "tippy.js": "^6.2.3",
     "typeface-lato": "^0.0.75"
   },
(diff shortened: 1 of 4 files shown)


Sl1nd:
There is some strange behaviour in changing the year:


Uploading Bildschirmvideo aufnehmen 2021-01-27 um 14.52.57.mov…

Sl1nd:
There is some strange behaviour on changing the Year in Monthpicker: 


https://user-images.githubusercontent.com/12165722/106001109-dab88080-60af-11eb-80bb-a305b36edee9.mov

janivo:
I'll have a look at it. Thanks @Sl1nd

pfecht (approved):
Always good to get rid auf some code :) Looks good to me, the behaviour changed. No odd behaviour for me.

The only strange thing is that the datepicker loses focus on clicking the next month (was the same before). We should use icon-buttons here and render our custom markup (if possible).

But this could be addressed separately.

janivo:
@pfecht: The focus thingy will be removed with #275 

Don't you have the same behaviour as @Sl1nd? For me, it behaves the same way shown in the video. 
Open Monthpicker -> Select Date -> Open it again -> Change year with the arrows at least twice -> See error

janivo:
fixed in #275

PR #273, opened 2021-01-26
feat(elements|ino-nav-item): add prop for sub text

janivo:
Adds a secondary text prop analogous to the ino-list-item

### DIFF
--- a/packages/elements/src/components.d.ts
+++ b/packages/elements/src/components.d.ts
@@ -775,13 +775,17 @@ export namespace Components {
     }
     interface InoNavItem {
         /**
-          * Styles the row in an activated style.  Use this for only one item and to mark it as permantently activated.
+          * Styles the row in an activated style.  Use this for only one item and to mark it as permanently activated.
          */
         "inoActivated"?: boolean;
         /**
           * Styles the row in a disabled style.
          */
         "inoDisabled"?: boolean;
+        /**
+          * The secondary text of this list item used in a two-lined list.
+         */
+        "inoSecondaryText"?: string;
         /**
           * The text of this list item.
          */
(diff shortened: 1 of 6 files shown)


packages/elements/src/components.d.ts
```diff
@@ -775,13 +775,17 @@ export namespace Components {
...
          */
         "inoActivated"?: boolean;
         /**
           * Styles the row in a disabled style.
          */
         "inoDisabled"?: boolean;
+        /**
+          * The secondary text of this list item used in a two-lined list.
+         */
+        "inoSecondaryText"?: string;
```
Sl1nd, line 788: the term `Secondary` implies that there is also a `pirmaryText`. Because of this and according to `inoText` i would recommend renaming  the property to `inoSubText`.

packages/elements/src/components/ino-nav-item/readme.md
```diff
@@ -38,11 +38,12 @@ document
...
+| Property           | Attribute            | Description                                                                                                | Type      | Default     |
+| ------------------ | -------------------- | ---------------------------------------------------------------------------------------------------------- | --------- | ----------- |
+| `inoActivated`     | `ino-activated`      | Styles the row in an activated style.  Use this for only one item and to mark it as permanently activated. | `boolean` | `false`     |
+| `inoDisabled`      | `ino-disabled`       | Styles the row in a disabled style.                                                                        | `boolean` | `false`     |
+| `inoSecondaryText` | `ino-secondary-text` | The secondary text of this list item used in a two-lined list.                                             | `string`  | `undefined` |
+| `inoText`          | `ino-text`           | The text of this list item.                                                                                | `string`  | `undefined` |
```
Sl1nd, line 48: In addition to the Readme it would be great to see its possibilites in inoNav Storybook  examples

Sl1nd:
For subText Nav-Items the Block isn´t vertically centered because of the fix height of the `mdc-list-item`
grafik

packages/elements/src/components.d.ts
janivo: Ye, I was thinking about that too. For consistency reasons, we should rename `inoText` to `inoPrimaryText` but I did not want to introduce a breaking change for this yet.

I'll change it to `ino-sub-text` for now.

packages/elements/src/components/ino-nav-item/readme.md
janivo: There actually is an example in the story of the nav-item. Should I add some more?

image

Sl1nd: Did't noticed. That's totally fine

Sl1nd (approved):

PR #275, opened 2021-01-26
refactor(elements|ino-datepicker): separate validation and configuration from component

### LINKED ISSUE #257: Refactor datepicker
The datepicker is pretty blown up and should be refactored for better code maintainability and readability.

## Some ideas
Think of which part may be better in separated files. For instance, the entire *Type-Specific options* could easily be moved to a `type-specific-options.ts`. Furthermore, the logic of how to validate the field could be moved to a `validation.ts`.

janivo:
Closes #257 

## Proposed Changes

- Move validation logic to `validation.ts`
- Move type-specific options to `type-specific-options.ts`
- Add some more test to ensure the functionality
- Remove focus workaround from the past
- Remove `updateFlatpickr` method to support typing

~~⚠️ This PR is currently based on #271, so I'll wait until that one is merged and rebase this one afterward.~~
As there are some problems with #271, I rebased this one onto the current master so this can be reviewed now.


janivo:
This can be reviewed now

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -272,71 +283,59 @@ export class Datepicker implements ComponentInterface {
...
       ignoredFocusElements: [],
       locale: getDatepickerLocale(this.el),
       onValueUpdate: (_, newValue) => this.valueChange.emit(newValue),
-      onChange: () => this.focusInputField(),
     };
 
-    const typeSpecificOptions: Partial<BaseOptions> = this.getTypeSpecificOptions();
+    const typeSpecificOptions: Partial<BaseOptions> = getTypeSpecificOption(
+      this.inoType,
+      {
```
pfecht, line 319: Why not passing the Element itself? The Props should be public vars so you don'T have to map all props

pfecht (changes requested):

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -348,130 +347,17 @@ export class Datepicker implements ComponentInterface {
...
-    if (this.flatpickr) {
-      this.flatpickr.set(option, value);
-    }
-  }
-
   private dispose() {
     if (this.flatpickr) {
       this.flatpickr.destroy();
-
-      if (this.isMonthPicker()) {
```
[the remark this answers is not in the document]
janivo: You are right. They were removed because the branch was based on another one and I forgot to revert it.

packages/elements/src/components/ino-datepicker/type-specific-options.ts
```diff
@@ -0,0 +1,101 @@
+import monthSelectPlugin from 'flatpickr/dist/plugins/monthSelect';
+import { BaseOptions } from 'flatpickr/dist/types/options';
+
+export type PickerTypeOptions = 'date' | 'month' | 'time' | 'datetime';
```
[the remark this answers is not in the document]
janivo: I had to use a workaround because you cannot extract keys from enums pre TS 4.1 (see [here](https://stackoverflow.com/questions/52393730/typescript-string-literal-union-type-from-enum)) but it's still an improvement

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
janivo: Which element do you mean? `this`? I have to remap them sooner or later because they are `ino`-prefixed

pfecht: Ah missed that points, Makes sense 👍

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -272,71 +283,59 @@ export class Datepicker implements ComponentInterface {
...
+      }
+    );
+
     const options = { ...sharedOptions, ...typeSpecificOptions };
 
     const target = this.el.querySelector('ino-input > div') as HTMLElement;
     this.flatpickr = flatpickr(target, options);
 
-    if (this.isMonthPicker()) {
+    if (this.inoType === 'month') {
```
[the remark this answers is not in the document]
janivo: Actually, I wanted to keep the `type-specific-options` free of the flatpickr instance because I think this might lead to worse readability. As a reader, I would not expect the `type-specific-options.ts` to manipulate the flatpickr instance.

What do you think?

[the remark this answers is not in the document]
janivo: I extracted as much as I could without using the flatpickr instance. Can you check again?

janivo:
This refactoring might have fixed the bug https://github.com/inovex/elements/issues/215 as well. Can you confirm this? @pfecht

silentHoo:
@Sl1nd mentioned [another bug](https://github.com/inovex/elements/issues/310) in the old datepicker component. We should check if this bug is fixed here as well.

pfecht:
What about the pending comment about `onCreate`? otherwised it looks good to me 👍

janivo:
@pfecht: Which comment do you mean exactly?

pfecht:
> @pfecht: Which comment do you mean exactly?

I didn't submit the review a couple of weeks ago 🙈

packages/elements/src/components/ino-datepicker/type-specific-options.ts
```diff
@@ -0,0 +1,127 @@
+import flatpickr from 'flatpickr';
+import monthSelectPlugin from 'flatpickr/dist/plugins/monthSelect';
+import { BaseOptions } from 'flatpickr/dist/types/options';
+
+export type PickerTypeKeys = typeof PickerTypes[keyof typeof PickerTypes];
+
+// Workaround until Stencil is using TS 4.1, use enum then
+// see => https://stackoverflow.com/questions/52393730/typescript-string-literal-union-type-from-enum
```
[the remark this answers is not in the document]
janivo: I've tried to use the newest Stencil version but it seems like the `typescript` dependency in the `elements-angular` project (which is at `4.0.3`) takes precedence over the `typescript` version defined in stencil's dev dependencies.
I will fix this in another PR.

janivo:
@pfecht: Good idea to use the `onReady` and `onDestroy` methods of the flatpickr, didn't think about that. 

I've implemented the changes you suggested and made some more changes because I realized that I'm actually implementing the factory pattern without thinking about it. It might seem like a lot but I mostly encapsulated all picker types to its own file and renamed some stuff. 

I would appreciate your review one (last?) time 🙏🏼

Sl1nd:
> @Sl1nd mentioned [another bug](https://github.com/inovex/elements/issues/310) in the old datepicker component. We should check if this bug is fixed here as well.

The issue is not fixed by this PR. Updating validation state is missing in 'min' and 'max' watcher. Should we integrate the [fixes](https://github.com/inovex/elements/issues/310) into this PR?

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -157,7 +167,7 @@ export class Datepicker implements ComponentInterface {
 
   @Watch('inoDefaultDate')
   inoDefaultDateChanged(defaultDate: string) {
-    this.updateFlatpickr('inoDefaultDate', defaultDate);
+    this.flatpickr?.set('defaultDate', defaultDate);
```
Sl1nd, line 170: I suggest the former implementation doesn't work for setting 'inoDefaultDate' and 'inoDateFormat'. I think it would be great to add these fixes to the changelog.

Sl1nd: https://github.com/inovex/elements/issues/340

pfecht (approved):
Looks good to me with the factory, great work! The only difference i have seen is in the naming of the fiels `picker/Datepicker`, `picker/Timepicker` which i would just name `picker/date` and `picker/time` etc. But just nitpicking

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
janivo: I would suggest to fix this in a separate PR

PR #277, opened 2021-01-27
refactor(elements): remove ino-leading/trailing properties

### LINKED ISSUE #252: Dont' use properties for conditional slots
In some cases, we use properties in order to render conditional slots. For instance, `ino-button`
https://github.com/inovex/elements/blob/bf6f0e53b5ca85230155d18ec2a21ec57d8ecc54/packages/elements/src/components/ino-button/ino-button.tsx#L70-L78

has two properties to render markup for icon two slots.

https://github.com/inovex/elements/blob/bf6f0e53b5ca85230155d18ec2a21ec57d8ecc54/packages/elements/src/components/ino-button/ino-button.tsx#L162-L166

https://github.com/inovex/elements/blob/bf6f0e53b5ca85230155d18ec2a21ec57d8ecc54/packages/elements/src/components/ino-button/ino-button.tsx#L174-L178

Do not use properties as work around to check whether a slot exists. Instead, check the content of a slot with
* `this.el.querySelectorAll("[slot=ino-secondary-icon]").length > 0;`

This reduces the number of properties and avoids duplicate code. Further, there should be some kind of "slot utility" to handle this behaviour consistently (with and without shadow DOM).

### Additional context
Similar behaviour should be reconsidered for:
* `ino-chip`
* `ino-input`
* `ino-list-item` (already implements this behaviour => move to utility)
* `ino-table`

janivo:
Closes #252 

## Proposed Changes

- Add utility function to check for slot content
- Refactor components that have conditional slots to use the new utility function
- Update stories and test where necessary

⚠️ This introduces a BREAKING CHANGE due to the removal of the aforementioned properties which leads to typescript errors in consumer projects.


janivo:
I will add some documentation for the slots

janivo:
This is ready for review

P1NHE4D (approved):

PR #281, opened 2021-02-03
style(elements|ino-menu): inherit border-radius

### LINKED ISSUE #231: ino-menu: Inherit border-radius
It would nice to customize the `border-radius` in the `ino-menu`.

I would like to define the `border-radius` like this:
```css
ino-menu {
  border-radius: 30px;
}
```

The alternative would be to dive deep into the children of `ino-menu`, select the correct container, and set the `border-radius` by myself. But this is an implementation detail and things like html-structure, classnames, etc. could change anytime. 

### Additional context
To solve this, just inherit the `border-radius` property to the correct container (in this case: `mdc-menu mdc-menu-surface mdc-menu-surface--open`). IMO a CSS-Var would be overkill.

See also: #232

P1NHE4D:
⚙️ **Changes:**
* adjusted the `border-radius` property of the `mdc-menu` class to inherit its value from `ino-menu`

🎟️ **Closes #231**

### DIFF
--- a/packages/elements/src/components/ino-menu/ino-menu.scss
+++ b/packages/elements/src/components/ino-menu/ino-menu.scss
@@ -4,6 +4,7 @@
 
 ino-menu {
   display: block;
+  border-radius: 4px;
 
   .mdc-menu-surface--anchor {
     border-radius: inherit;
--- a/packages/elements/src/components/ino-menu/ino-menu.scss
+++ b/packages/elements/src/components/ino-menu/ino-menu.scss
@@ -4,7 +4,6 @@
 
 ino-menu {
   display: block;
-  border-radius: 4px;
 
   .mdc-menu-surface--anchor {
     border-radius: inherit;
--- a/packages/elements/src/components/ino-menu/ino-menu.scss
+++ b/packages/elements/src/components/ino-menu/ino-menu.scss
@@ -4,4 +4,13 @@
 
 ino-menu {
   display: block;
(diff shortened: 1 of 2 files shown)


packages/elements/src/components/ino-menu/ino-menu.scss
```diff
@@ -4,4 +4,13 @@
 
 ino-menu {
   display: block;
+  border-radius: 4px;
```
janivo, line 7: The `4px` overwrites the value I provide. It does not change the border-radius with the following code:
```css
ino-menu {
  border-radius: 0;
}
```

janivo (changes requested):

packages/elements/src/components/ino-menu/ino-menu.scss
P1NHE4D: You should now be able to set it to a different value. I removed the property from the stylesheet.

janivo: Ok, my bad here.
I thought that it behaves dynamically e.g. I can change the `border-radius` to `99px` in the DEV-Tools and it updates instantly. Turns out that it's static and was working before and I did not use it correctly.
I'll revert your last commit and approve it!

P1NHE4D: Oh good to know! 😄 Thanks for the heads up!

janivo (approved):

PR #282, opened 2021-02-03
style(elements|ino-card): inherit width and height

### LINKED ISSUE #232: ino-card: Inherit common attributes
Same as in #231 but with the `ino-card` component.

Additional attributes which should be inherited:
- `height`
- `width`
- `margin` (maybe?)
- `padding` (maybe?)

**Possible solution**:
```css
div {
  width: inherit;
}
```

P1NHE4D:
⚙️ **Changes:**
* the `width`, `height`, and `border-radius` properties inherit their values from `ino-card`
* some additional changes to the styling to reflect dimensional changes correctly

⚠️ **Remarks:**
* the `margin` and `padding` properties do not inherit their values from `ino-card` since setting the corresponding properties of the `ino-card` changes the alignment of the `ino-card` component itself with respect to its environment

🎟️ **Closes #232**

### DIFF
--- a/packages/elements/src/components/ino-card/ino-card.scss
+++ b/packages/elements/src/components/ino-card/ino-card.scss
@@ -24,6 +24,7 @@ ino-card {
 ino-card {
   display: block;
   position: relative;
+  border-radius: 20px;
 
   .mdc-card {
     @include card.outline(theme.color(light, base), 1px);
--- a/packages/elements/src/components/ino-card/ino-card.scss
+++ b/packages/elements/src/components/ino-card/ino-card.scss
@@ -16,7 +16,7 @@ ino-card {
   --card-background-color: var(
     --ino-card-background-color,
     --ino-global-primary-color,
-    #{theme.color(primary, base)}
+    #{theme.color(primary, $tone: 'base')}
   );
   --card-color: var(--ino-card-color, #fff);
 }
@@ -26,9 +26,18 @@ ino-card {
   position: relative;
 
   .mdc-card {
-    @include card.outline(theme.color(light), 1px);
+    @include card.outline(theme.color(light, $tone: 'base'), 1px);
     box-shadow: none;
     border-radius: 20px;
+    height: inherit;
+    width: inherit;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+
+    &::after {
+      content: unset;
+    }
   }
 
   .ino-card--selected.mdc-card {
--- a/packages/elements/src/components/ino-card/ino-card.scss
+++ b/packages/elements/src/components/ino-card/ino-card.scss
@@ -28,7 +28,7 @@ ino-card {
   .mdc-card {
     @include card.outline(theme.color(light, base), 1px);
     box-shadow: none;
-    border-radius: 20px;
+    border-radius: inherit;
     height: inherit;
     width: inherit;
     display: flex;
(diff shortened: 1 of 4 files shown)


packages/elements/src/components/ino-card/ino-card.scss
```diff
@@ -16,7 +16,7 @@ ino-card {
   --card-background-color: var(
     --ino-card-background-color,
     --ino-global-primary-color,
-    #{theme.color(primary, base)}
+    #{theme.color(primary, $tone: 'base')}
```
janivo, line 19: I think you can just write `theme.color(primary, base)`, can't you?

packages/elements/src/components/ino-card/ino-card.scss
```diff
@@ -26,9 +26,18 @@ ino-card {
   position: relative;
 
   .mdc-card {
-    @include card.outline(theme.color(light), 1px);
+    @include card.outline(theme.color(light, $tone: 'base'), 1px);
```
janivo, line 29: same as above

packages/elements/src/components/ino-card/ino-card.tsx
```diff
@@ -40,7 +40,7 @@ export class Card implements ComponentInterface {
     return (
       <Host>
         <div class={classList}>
-          <div>
+          <div style={{"display": "contents"}}>
```
janivo, line 43: We should be using a class instead of this inline-style

packages/elements/src/components/ino-card/readme.md
```diff
@@ -59,14 +59,6 @@ class MyComponent extends Component {
 | `inoSelected`         | `ino-selected`          | Selects the card and displays a check mark icon on top of the card | `boolean` | `false` |
 
 
-## CSS Custom Properties
```
janivo, line 62: This seems to be happening while running the `yarn start` script. Run `yarn build` in `packages/elements` to restore this table

janivo (changes requested):
Could you also inherit `border-radius`? I forgot to mention this in the ticket

packages/elements/src/components/ino-card/ino-card.scss
P1NHE4D: I think so, yes. I only changed it because my IDE was complaining for some reason...

packages/elements/src/components/ino-card/ino-card.tsx
P1NHE4D: ✅

packages/elements/src/components/ino-card/readme.md
P1NHE4D: Done ✅

packages/storybook/src/stories/ino-card/ino-card.scss
```diff
@@ -63,4 +63,5 @@
 
 ino-card {
   width: 300px;
+  border-radius: 20px;
```
janivo, line 66: This should be moved to the component itself just like in #281.
Otherwise, this breaks the current design

janivo (changes requested):

packages/storybook/src/stories/ino-card/ino-card.scss
P1NHE4D: Will do ✅

janivo (approved):

Issue #288, opened 2021-02-15
ino-menu: Unable to change positioning


janivo:
### Is your feature request related to a problem? Please describe.

By default, the ino-menu component positions itself. If this functionality does not really work (e.g. outside the viewport), there should be an option to set it manually.

![image](https://user-images.githubusercontent.com/22963121/107950150-a17c7d80-6f96-11eb-88b4-260e9d347561.png)

### Describe the solution you'd like

A property that takes a corner (e.g. `TOP_LEFT`) and positions the menu surface at that position.

### Describe alternatives you've considered

I tried setting the menu in angular by myself but the mdc-menu seems to apply some inline styles which are hard to overwrite.

PR #289, opened 2021-02-16
feat(elements|ino-popover): add property to manually show/hide

janivo:
Relates to #288 

In order to implement a better menu, it should be based on the popover and not on the MDC-Menu.
This PR provides functionalities to manage the state of the popover.

## Proposed Changes

- Add new `ino-show` property which controls the state of the popover
- Add `clickOutside` event to simplify the state management
- Update Storybook entry of the `ino-popover`

### DIFF
--- a/packages/elements/src/components/ino-popover/ino-popover.tsx
+++ b/packages/elements/src/components/ino-popover/ino-popover.tsx
@@ -1,8 +1,8 @@
 import {
   Component,
   ComponentInterface,
-  Element,
   Event,
+  Element,
   EventEmitter,
   h,
   Host,
@@ -85,14 +85,8 @@ export class Popover implements ComponentInterface {
 
   @Watch('inoShow')
   inoShowChanged(show: boolean) {
-    console.log('Show: ', show);
     show ? this.tippyInstance.show() : this.tippyInstance.hide();
   }
-
-  /**
-   * Emits the visibility of the popover on change (true if shown, false if hidden).
-   */
-  @Event() visibilityChanged!: EventEmitter<boolean>;
 
   // Lifecycle
 
@@ -100,7 +94,11 @@ export class Popover implements ComponentInterface {
     this.create();
   }
 
-  // Private methods
+  /**
+   * Emits when an element which is not part of the popover is clicked.
+   * Should be used if you control the state of the popover.
+   */
+  @Event() clickOutside: EventEmitter<void>;
 
   private create() {
     this.tippyInstance?.destroy();
@@ -115,30 +113,22 @@ export class Popover implements ComponentInterface {
       );
     }
 
+    const isControlled = this.inoShow !== undefined;
+
     const options: Partial<Props> = {
       allowHTML: true,
       appendTo: this.el.parentElement,
       content: this.el,
       duration: 100,
+      hideOnClick: !isControlled,
       placement: this.inoPlacement,
-      trigger: this.inoTrigger,
-      interactive: this.inoInteractive,
+      trigger: isControlled ? 'manual' : this.inoTrigger,
+      interactive: isControlled || this.inoInteractive,
+      showOnCreate: isControlled && this.inoShow,
+      onClickOutside: () => this.clickOutside.emit(),
     };
 
-    if (this.inoShow !== undefined) {
-      console.log('adding handlers');
-      options.onShow = (_) => this.handleVisibility(true);
-      options.onHide = (_) => this.handleVisibility(false);
-      options.showOnCreate = this.inoShow;
-    }
-
     this.tippyInstance = TippyJS(target, options);
-  }
-
-  handleVisibility(show: boolean): false | void {
-    console.log('Visibility: ', show, this);
-    this.visibilityChanged.emit(show);
-    return false;
   }
 
   render() {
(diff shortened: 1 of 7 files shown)


packages/elements/src/components/ino-popover/ino-popover.tsx
```diff
@@ -81,26 +94,41 @@ export class Popover implements ComponentInterface {
...
+      console.warn(
+        `The element with the id '${this.inoFor}' could not be found.`
+      );
+    }
+
+    const isControlled = this.inoShow !== undefined;
+
+    const options: Partial<Props> = {
       allowHTML: true,
       appendTo: this.el.parentElement,
```
pfecht, line 120: Does this still make sense? I changed it to `'parent'` in order to place it as parent of the target, not the popover element.

packages/elements/src/components/ino-popover/ino-popover.tsx
```diff
@@ -81,26 +94,41 @@ export class Popover implements ComponentInterface {
...
       ? document.getElementById(this.inoFor)
       : this.el.parentElement;
 
-    const options = {
+    if (!target && this.inoFor) {
+      console.warn(
+        `The element with the id '${this.inoFor}' could not be found.`
+      );
+    }
+
```
pfecht, line 115: Maybe also a good moment to resolve the jumping content problem

```ts
// Add the display property on creation to avoid jumping content
// see https://atomiks.github.io/tippyjs/v6/html-content/#element
this.el.style.display = 'block';
```

and 

```css
ino-popover {
  display: none;
}

File: packages/elements/src/components/ino-popover/ino-popover.tsx [hunk 3]
```diff
@@ -81,26 +94,41 @@ export class Popover implements ComponentInterface {
...
       duration: 100,
+      hideOnClick: !isControlled,
       placement: this.inoPlacement,
-      trigger: this.inoTrigger,
-      interactive: this.inoInteractive,
+      trigger: isControlled ? 'manual' : this.inoTrigger,
+      interactive: isControlled || this.inoInteractive,
+      showOnCreate: isControlled && this.inoShow,
+      onClickOutside: () => this.clickOutside.emit(),
     };
```
pfecht, line 129: For more customization, won't you add a distance property to configure the tippys offset? 

` options['offset'] = [0, this.inoDistance];`.

Default distance is 10px which does not apply to all usecases

packages/elements/src/components/ino-popover/ino-popover.tsx
```diff
@@ -81,26 +94,41 @@ export class Popover implements ComponentInterface {
     this.create();
   }
 
-  // Private methods
+  /**
+   * Emits when an element which is not part of the popover is clicked.
+   * Should be used if you control the state of the popover.
+   */
+  @Event() clickOutside: EventEmitter<void>;
```
pfecht, line 101: Isn't this a bottleneck? Like this we can only use the popover in controlled mode using a click. What happens if I define a hover trigger, f.e?

packages/elements/src/components/ino-popover/readme.md
```diff
@@ -1,20 +1,84 @@
...
+          This popover will show as soon as the user hovers the button above
+        </InoPopover>
+      </div>
+    );
+  }
+}
+```
+_Controlled_
+
+Or you use the `ino-open` property to show/hide the popover by yourself.
```
pfecht, line 49: `ino-show`

packages/elements/src/components/ino-popover/readme.md
```diff
@@ -1,20 +1,84 @@
...
+        </InoPopover>
+      </div>
+    );
+  }
+}
+```
+_Controlled_
+
+Or you use the `ino-open` property to show/hide the popover by yourself. 
+This is helpful if you want to implement custom logic when the popover should be shown or hid.
```
pfecht, line 50: `hid` => `hidden`

packages/storybook/src/stories/ino-popover/ino-popover.stories.js
```diff
@@ -70,5 +105,11 @@ export const DefaultUsage = () => /*html*/ `
             <ino-button>You can click me without closing this popover!</ino-button>
         </div>
       </ino-popover>
+      </div>
+      <h4>Controlled Popover</h4>
+      <ino-checkbox id="controlled-checkbox" checked="true">Uncheck to hide / check to show</ino-checkbox>
```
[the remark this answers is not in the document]
janivo: Agreed but this way it's easier to select the popover as it is already in the DOM.

packages/elements/src/components/ino-popover/ino-popover.tsx
janivo: Yea, if we want to be really consistent, there also should be events like `hoverOutside`, ...
In this case, I just forwarded the tippy API (`onClickOutside`) and thought it is nice to have if you want to implement custom logic.

packages/elements/src/components/ino-popover/ino-popover.tsx
janivo: Nice idea but I would rather do it in a separate issue so the new 'feature' will be visible to the consumer in our changelog.

packages/elements/src/components/ino-popover/ino-popover.tsx
janivo: https://github.com/inovex/elements/issues/291

packages/elements/src/components/ino-popover/ino-popover.tsx
janivo: https://github.com/inovex/elements/issues/291

janivo:
updated version found in https://github.com/inovex/elements/pull/292

PR #299, opened 2021-03-01
feat(elements|ino-popover): add prop `ino-distance` to define offset

### LINKED ISSUE #291: ino-popover: Add custom offset property
The `<ino-popover>` uses the [tippy.js library](https://atomiks.github.io/tippyjs/) under the hood which provides us a property to define a custom offset: https://atomiks.github.io/tippyjs/v6/all-props/#offset

The default distance is 10px which does not apply to all use-cases.
We should implement it as a property (maybe `ino-offset`?).

_Originally posted by @pfecht in https://github.com/inovex/elements/pull/289#discussion_r576987138_

janivo:
Closes #291

## Proposed Changes

- Forward `ino-distance` prop to tippy's `offset` option
- Change watcher to not recreate tippy on every minor change


packages/storybook/src/stories/ino-popover/ino-popover.stories.js
```diff
@@ -1,4 +1,4 @@
-import { boolean, select } from '@storybook/addon-knobs';
+import { boolean, number, select } from '@storybook/addon-knobs';
```
pfecht, line 1: The `console.log` from line 10 is a left over from the previous PR

packages/elements/src/components/ino-popover/ino-popover.tsx
```diff
@@ -142,6 +164,7 @@ export class Popover implements ComponentInterface {
       placement: this.inoPlacement,
       trigger: this.inoTrigger,
       interactive: this.inoInteractive,
+      offset: [0, this.inoDistance],
```
pfecht, line 167: Cannot add a review for unchanged lines :( Another left over from the previous MR:
In l. 209 there is a second `ino-popover__content` which is a duplicate of the content a few lines above. In my opinion the second content could be removed.

pfecht (approved):

packages/elements/src/components/ino-popover/ino-popover.tsx
janivo: You mean this:
https://github.com/inovex/elements/blob/ce8569a1b5ef384d2dfb803133a7c924766e0c43/packages/elements/src/components/ino-popover/ino-popover.tsx#L185-L189

Should be replaced with:
```html
<div class="ino-tooltip__inner">
  <slot></slot>
</div>
```
?

janivo: Done ✅

pfecht (approved):

PR #306, opened 2021-03-12
feat(elements|*): add focus and blur methods

pfecht:
Adds methods to set `focus` and `blur` on native input elements.

## Proposed Changes

-


packages/elements/src/components/ino-input/ino-input.tsx
```diff
@@ -236,6 +236,24 @@ export class Input implements ComponentInterface {
     return this.nativeInputEl;
   }
 
+  /**
+   * Sets focus on the native `input`. 
+   * Use this method instead of the global `input.focus()`.
+   */
+   @Method()
+   async setFocus() {
```
janivo, line 244: Could you explain why this method should be used over the native `focus()` function?
The focus function seems to be doing the same thing as:
```ts
@Listen('focus')
focusListener() {
  this.setFocus();
}
```

PR #307, opened 2021-03-12
refactor(elements): use listen decorator

### LINKED ISSUE #249: Refactor component listeners
### Describe the bug
Some components define custom event listeners and add them to the element instead of using the stencil Listen()-decorator (e.g. ino-snackbar, ino-dialog, ...). In addition some of those event listeners are not removed within the disconnected lifecycle of the respective component.

### To Reproduce
1. Search for 'addEventListener' in the project 
2. Have a look at the component tsx files in which the event listeners are attached to the element itself
3. In addition check, if the event listeners are removed within the disconnected callback 

### Expected behavior
- the stencil Listen()-decorator is used whenever possible, because it handles the listeners' detachment
- all event listeners are removed in the disconnected lifecycle of the component

### Possible Solution
- substitute custom event listeners by using the Listen()-decorator (if applicable)
- remove custom event listeners within the disconnectedCallback lifecycle

P1NHE4D:
⚙️ **Changes:**
* Removed the event listeners from the `ino-dialog` component since the events were never dispatched due to the missing `data-mdc-dialog-action` attribute (see [mdc-dialog documentation](https://github.com/material-components/material-components-web/tree/master/packages/mdc-dialog#full-screen-dialog))
* Added stencil's `Listen()` decorator to the `ino-tab` and `ino-tab-bar` components

⚠️ **The `Listen()` decorator can not be used in the following components:**
| Component | Reason |
| ----------- | -------- |
| `ino-nav-drawer` | Event listener is attached to a specific target. The only targets supported by `Listen()` are `window`, `document`, and `body`. |
| `ino-chip-set` | Event listener is only attached in specific cases. |
| `ino-datepicker` | Event listener is attached to a specific target. |
| `ino-file-input` | Listens to multiple events at the same time. What is more, some events are only attached if they're supported by the browser. |
| `ino-radio-group` | Listeners are attached to specific targets. |
| `ino-segment-button` | Listeners are only attached if the segment button belongs to a group. |
| `ino-tooltip` | Listeners are attached to specific targets. |

🎟️ **Closes #249**

### DIFF
--- a/packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
+++ b/packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
@@ -9,7 +9,7 @@ import {
   Prop,
   Watch,
   h, Listen
-} from "@stencil/core";
+} from '@stencil/core';
 
 @Component({
   tag: 'ino-tab-bar',
--- a/packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
+++ b/packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
@@ -8,8 +8,8 @@ import {
   Host,
   Prop,
   Watch,
-  h,
-} from '@stencil/core';
+  h, Listen
+} from "@stencil/core";
 
 @Component({
   tag: 'ino-tab-bar',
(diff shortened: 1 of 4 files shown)


packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
```diff
@@ -8,8 +8,8 @@ import {
   Host,
   Prop,
   Watch,
-  h,
-} from '@stencil/core';
+  h, Listen
+} from "@stencil/core";
```
janivo, line 12: The `'...'` should stay

janivo (changes requested):
Just a really small change

packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
P1NHE4D: fixed ✅

janivo (approved):

PR #309, opened 2021-03-24
refactor(elements): replace attribute selectors

### LINKED ISSUE #243: Replace attribute selectors with classes
At some points, we use attribute selectors to style our elements based on properties. For example: https://github.com/inovex/elements/blob/9db4d4f0285bc71bc8c7d5a60586b20b78c2820e/packages/elements/src/components/ino-button/ino-button.scss#L233-L242

We should replace these with classes. This brings the following advantages:
- Code will be shorter because we don't have to check for strings (see example above)
- Increased consistency in stylesheets
- Angular's reflection system won't collide with our styles (see https://github.com/inovex/elements/pull/240)

janivo:
Closes #243

## Proposed Changes

- Replaces all attribute selectors
- Some small improvements to the storybook


P1NHE4D (changes requested):

P1NHE4D:
@janivo Could you resolve the conflicts? 🙂

janivo:
@P1NHE4D: thanks for the heads up. I rebased it. Only things that I had to resolve were the `ino-popover` and `ino-tooltip`.

P1NHE4D (approved):

PR #311, opened 2021-03-26
style(elements|ino-card): add css variables for the background color and border

### LINKED ISSUE #308: ino-card: Add a custom-property to change background color
https://github.com/inovex/elements/blob/ea03f3dc242beb379e70e19406eb2d7e11429086/packages/elements/src/components/ino-card/ino-card.scss#L16-L21

These CSS-Vars are currently only being used for the icon color of the card.
As a consumer, I would expect a CSS-Var which can change the background color of the whole card as well (e.g. for focus or hover styles).

### LINKED ISSUE #334: ino-card: Blue variant of a selected card
In our self onboarding, the selected state on the ino-card is currently displayed with a blue circle (see attachment). This should be integrated into the elements, as it fits very well with the design and would then be consistent with the ino-checkbox.

image

P1NHE4D:
✨ **Features:**
* The background color and border of the card can now be customised using CSS variables

🎟️ **Closes #308** 
🎟️ **Closes #334**

### DIFF
--- a/packages/elements/src/components/ino-card/ino-card.scss
+++ b/packages/elements/src/components/ino-card/ino-card.scss
@@ -21,7 +21,7 @@ ino-card {
   --card-background-color--active: var(--ino-card-background-color--active, #fff);
   --card-background-color--focus: var(--ino-card-background-color--focus, #fff);
   --card-background-color--hover: var(--ino-card-background-color--hover, #fff);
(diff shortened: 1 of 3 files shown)


P1NHE4D:
> The border of the cards look kind of different than before:
> 
> ## Before
> image
> 
> ## After
> image
> 
> Could you check on why this is happening?

Should work now

janivo (approved):

PR #313, opened 2021-03-26
fix(elements|ino-datepicker): validate on min/max date changes

### LINKED ISSUE #310: ino-datepicker: set validation state on min/max value change
### Describe the bug

Min an Max Date changes do not update validation state. 

### Expected behaviour

If the minimum or maximum date changes the current value should be validated.

### Possible Solution

Additionally watch min/max date changes and trigger validation logic.

Sl1nd:
Closes #310 

## Proposed Changes

- Validate on min/max date changes


janivo:
@Sl1nd: Nice 👍🏼  

I would wait for #275 to be merged and then apply these changes if that's okay for you.

janivo:
Hey @Sl1nd, could you redo your changes with the new master? We did just merge a huge datepicker update. Should be easy to transfer these changes to the new master.

janivo:
One thing I noticed is that the already selected date will be erased if the `minDate` or `maxDate` of the flatpickr changes and the current value is out of the new range. You can test it this way:
1. Select a date within the provided `min`/`max` (e.g. `value="12.04.2021"`)
2. Change the `min`/`max` prop of the `ino-datepicker` to a new date where the selected date from Step 1 is out of range (e.g. `max="11.04.2021"`)
3. See that the `value` prop will be unset

I think that this behavior is reasonable and would consider it as expected. What do you think @Sl1nd?

Sl1nd:
> One thing I noticed is that the already selected date will be erased if the `minDate` or `maxDate` of the flatpickr changes and the current value is out of the new range. You can test it this way:
> 
>     1. Select a date within the provided `min`/`max` (e.g. `value="12.04.2021"`)
> 
>     2. Change the `min`/`max` prop of the `ino-datepicker` to a new date where the selected date from Step 1 is out of range (e.g. `max="11.04.2021"`)
> 
>     3. See that the `value` prop will be unset
> 
> 
> I think that this behavior is reasonable and would consider it as expected. What do you think @Sl1nd?

I suggest this behavior relates to the implementation of flatpickr? As also the overlay selection of dates  outta range is disabled i suppose this is an expected behaviour. Maybe it would be more confusing if the value is still present and not selected in overlay anymore.

Sl1nd:
@janivo  May be merged

janivo (approved):

PR #317, opened 2021-04-02
feat(elements|ino-select): add slot for leading icon

### LINKED ISSUE #233: ino-select: Add support for icons
I need to place an icon next to my `ino-select`. As the `ino-select` and the `ino-input` look very similar, it only makes sense to add support for icons to the select as well.

I would like to provide an `ino-icon` or any external icon per slot to the `ino-select`. See `ino-input` for example.

P1NHE4D:
✨ **Features:**
* added `ino-icon-leading` slot to `ino-select` component

🎟️ **Closes #233**

### DIFF
--- a/packages/elements/src/components/ino-select/ino-select.scss
+++ b/packages/elements/src/components/ino-select/ino-select.scss
@@ -115,6 +115,13 @@ ino-select {
     .mdc-floating-label.mdc-floating-label--float-above {
       top: 25px;
     }
+
+    &[class*='--with-leading-icon'] {
+      .mdc-floating-label,
+      .mdc-floating-label--float-above {
+        left: 40px;
+      }
+    }
   }
 
   .mdc-select--outlined {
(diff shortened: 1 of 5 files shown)


janivo (changes requested):
image
The Icon seems kinda huge. We should use the same styles applied in the `ino-input` component. Maybe we can extract similar styles in a mixin? We could create a `mixin.scss` in `/base` and use those mixin in both components.

P1NHE4D:
> image
> The Icon seems kinda huge. We should use the same styles applied in the `ino-input` component. Maybe we can extract similar styles in a mixin? We could create a `mixin.scss` in `/base` and use those mixin in both components.

I initially made the icon smaller, but it looked a bit odd next to the rather big drop-down selector, especially with a floating label.

janivo (changes requested):
I added a story as an example and noticed that the label is not positioned properly yet:
image

janivo (approved):

PR #336, opened 2021-04-07
fix(elements|ino-nav-drawer): emit `openChange` event on `esc` press

### LINKED ISSUE #328: ino-nav-drawer: open state on escape (modal variant)
If you open and close the ino-nav-drawer with the "esc" key the open state is not set correctly. Here you have to click 2x on the toggle button to open the nav-drawer again.

**Acceptance criteria**
* openChange is also thrown with ESC.
* openChange is renamed to openChanged (BREAKING!)

P1NHE4D:
⚙️ **Changes:**
* `openChange` event is emitted on pressing `esc`

🎟️ **Closes #328**

### DIFF
--- a/packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
+++ b/packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
@@ -67,6 +67,8 @@ export class NavDrawer implements ComponentInterface {
       this.drawerInstance.open = this.open || false;
     }
 
+    this.el.shadowRoot.querySelector('.mdc-drawer').addEventListener('MDCDrawer:closed', this.closeDrawer);
+
     this.el.shadowRoot
       .querySelector('.mdc-drawer__toggle')
       .addEventListener('click', this.toggleDrawer);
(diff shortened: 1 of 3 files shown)


packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
```diff
@@ -67,6 +67,10 @@ export class NavDrawer implements ComponentInterface {
       this.drawerInstance.open = this.open || false;
     }
 
+    this.el.shadowRoot
```
janivo, line 70: Instead of querying for the element, I would suggest to use it's reference like this:
```jsx
<aside class={classDrawer} ref={(el) => this.drawerEl = el}>
```

packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
```diff
@@ -67,6 +67,10 @@ export class NavDrawer implements ComponentInterface {
...
+      .querySelector('.mdc-drawer')
+      .addEventListener('MDCDrawer:closed', this.closeDrawer);
```
janivo, line 72: We need to remove the listener on unloading.

janivo (changes requested):

packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
P1NHE4D: Done ✅ I also directly attached the event listener for the `click` event to the `ino-icon-button`.

packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
P1NHE4D: Done ✅ I don't know why I always forget to remove the event listener...

janivo (approved):

PR #338, opened 2021-04-07
feat(elements): allow lists with both single and double lines

### LINKED ISSUE #284: InoList: Allow single- and twolined items

Currently it is not possible to use mixed `ino-lists` with single and two-lined `ino-list-items`. Either the singlelined list-item  or the twolined list-item isn't vertically centered.

### Describe possible solutions

Maybe there could be a third option called e.g. `mixed-lines` that allows to use both single- and twolined items.

### Additional context
Often Navigation items may be used with both single and twolined nav-items.

grafik

grafik

P1NHE4D:
✨ **Features:**
*  lists can now have items with either one or two lines

🎟️ **Closes #284**

### DIFF
--- a/packages/elements-angular/elements/src/directives/proxies.ts
+++ b/packages/elements-angular/elements/src/directives/proxies.ts
@@ -460,13 +460,13 @@ export class InoLabel {
 
 export declare interface InoList extends Components.InoList {}
 @ProxyCmp({
-  inputs: ['inoAvatar', 'inoDense', 'inoMixedLines', 'inoTwoLines']
+  inputs: ['inoAvatar', 'inoDense', 'inoTwoLines']
 })
 @Component({
   selector: 'ino-list',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
-  inputs: ['inoAvatar', 'inoDense', 'inoMixedLines', 'inoTwoLines']
+  inputs: ['inoAvatar', 'inoDense', 'inoTwoLines']
 })
 export class InoList {
   protected el: HTMLElement;
--- a/packages/elements-angular/elements/src/directives/proxies.ts
+++ b/packages/elements-angular/elements/src/directives/proxies.ts
@@ -460,13 +460,13 @@ export class InoLabel {
 
 export declare interface InoList extends Components.InoList {}
 @ProxyCmp({
-  inputs: ['inoAvatar', 'inoDense', 'inoTwoLines']
+  inputs: ['inoAvatar', 'inoDense', 'inoMixedLines', 'inoTwoLines']
 })
 @Component({
   selector: 'ino-list',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
-  inputs: ['inoAvatar', 'inoDense', 'inoTwoLines']
+  inputs: ['inoAvatar', 'inoDense', 'inoMixedLines', 'inoTwoLines']
 })
 export class InoList {
   protected el: HTMLElement;
--- a/packages/elements/src/components/ino-list/ino-list.scss
+++ b/packages/elements/src/components/ino-list/ino-list.scss
@@ -8,4 +8,12 @@ ino-list {
   .mdc-list--avatar-list .mdc-list-item__graphic {
     overflow: hidden;
   }
+
+  .mdc-list--mixed-line .mdc-list-item {
+    height: 64px;
+  }
+
+  .mdc-list--mixed-line.mdc-list--dense .mdc-list-item {
+    height: 60px;
+  }
 }
--- a/packages/elements/src/components/ino-list/ino-list.scss
+++ b/packages/elements/src/components/ino-list/ino-list.scss
@@ -8,12 +8,4 @@ ino-list {
   .mdc-list--avatar-list .mdc-list-item__graphic {
     overflow: hidden;
   }
-
-  .mdc-list--mixed-line .mdc-list-item {
-    height: 64px;
-  }
-
-  .mdc-list--mixed-line.mdc-list--dense .mdc-list-item {
-    height: 60px;
-  }
 }
--- a/packages/elements/src/components.d.ts
+++ b/packages/elements/src/components.d.ts
@@ -686,10 +686,6 @@ export namespace Components {
          */
         "inoDense"?: boolean;
         /**
-          * Marks this element as a list with items having either one or two lines.
-         */
-        "inoMixedLines"?: boolean;
-        /**
           * Marks this element as list with items having two lines.
          */
         "inoTwoLines"?: boolean;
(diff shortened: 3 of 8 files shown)


packages/elements-angular/elements/src/directives/proxies.ts
```diff
@@ -460,13 +460,13 @@ export class InoLabel {
...
 @ProxyCmp({
-  inputs: ['inoAvatar', 'inoDense', 'inoTwoLines']
+  inputs: ['inoAvatar', 'inoDense', 'inoMixedLines', 'inoTwoLines']
 })
 @Component({
   selector: 'ino-list',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
-  inputs: ['inoAvatar', 'inoDense', 'inoTwoLines']
+  inputs: ['inoAvatar', 'inoDense', 'inoMixedLines', 'inoTwoLines']
```
JCofman, line 469: Hi @P1NHE4D  I am not sure but could we may achieve the same effect without using the `inoMixedLines` prop overall :) ?

packages/elements/src/components/ino-list/ino-list.scss
```diff
@@ -8,4 +8,12 @@ ino-list {
   .mdc-list--avatar-list .mdc-list-item__graphic {
     overflow: hidden;
   }
+
+  .mdc-list--mixed-line .mdc-list-item {
```
JCofman, line 12: I think we could remove these extra classes and overwrite this

```
  .mdc-list--two-line .mdc-list-item__text {
    align-self: center;
  }
```

JCofman (changes requested):

packages/elements-angular/elements/src/directives/proxies.ts
P1NHE4D: I thought so as well, but it would require to either set the height of the list items using JS or increase the default height for list items with a single line only. If I add a list-item with two lines and set `inoTwoLines` to `false`, the height of the list item would be too narrow. Nonetheless, if `inoTwoLines == true`, the list-items are padded to ensure that each list item can fit two lines, regardless of the actual number of lines.

Without `inoMixedLines`, we would have to check if the list contains any items with more than one line. If this is the case, the height of the list items needs to be increased in order to make sure that the list items have the same height and to prevent the item with two lines to be cut off. At least that's how I would solve it.

JCofman: Thanks a lot for your clarification :) @P1NHE4D  

I played around a bit and I think I might found a better solution to the problem. We could remove some default MDC CSS and could handle two and one line lists without having to set `inoTwolines` or `inoMixedLines`.

This would make the API more flexible and more simple to understand and I think wouldn't have any breaking change. We could easily get rid of two props but handling both use cases.

```diff
/* ino-list-item.scss */

@use 'base/mdc-customize';
@use '@material/list/mdc-list';
@use 'base/theme';

ino-list-item {
  /**
   * @prop --ino-list-item-selected-color: Text color of a selected list item
   * @prop --ino-list-item-selected-background-color: Background color of a selected list item
   * @prop --ino-list-item-selected-background-color-hover: Background color of a selected list item on hover
   * @prop --ino-list-item-selected-background-color-active: Background color of a selected list item if active
   * @prop --ino-list-item-selected-background-color-focus: Background color of a selected list item if focused
   * @prop --ino-list-item-deselected-color: Text color of a deselected list item
   * @prop --ino-list-item-deselected-background-color: Background color of a deselected list item
   * @prop --ino-list-item-deselected-background-color-hover: Background color of a deselected list item on hover
   * @prop --ino-list-item-deselected-background-color-active: Background color of a deselected list item if active
   * @prop --ino-list-item-deselected-background-color-focus: Background color of a deselected list item if focused
   */
  --list-item-selected-color: var(
    --ino-list-item-selected-color,
    #{theme.color(primary)}
  );
  --list-item-selected-background-color: var(
    --ino-list-item-selected-background-color,
    #{rgba(theme.color(primary, base), 0.05)}
  );
  --list-item-selected-background-color-hover: var(
    --ino-list-item-selected-background-color-hover,
    #{rgba(theme.color(primary, base), 0.1)}
  );
  --list-item-selected-background-color-active: var(
    --ino-list-item-selected-background-color-active,
    #{rgba(theme.color(primary, base), 0.3)}
  );
  --list-item-selected-background-color-focus: var(
    --ino-list-item-selected-background-color-focus,
    #{rgba(theme.color(primary, base), 0.15)}
  );
  --list-item-deselected-color: var(--ino-list-item-deselected-color);
  --list-item-deselected-background-color: var(
    --ino-list-item-deselected-background-color
  );
  --list-item-deselected-background-color-hover: var(
    --ino-list-item-deselected-background-color-hover
  );
  --list-item-deselected-background-color-active: var(
    --ino-list-item-deselected-background-color-active
  );
  --list-item-deselected-background-color-focus: var(
    --ino-list-item-deselected-background-color-focus
  );
}

ino-list-item {
  display: block;

  .ino-list-item__icon {
    font-size: 24px;
  }

  .mdc-list-item__graphic:empty {
    display: none;
  }

  ino-img .ino-img__image {
    width: 100%;
    height: 100%;
  }

  .mdc-list-item {
    color: var(--list-item-deselected-color);
    background-color: var(--list-item-deselected-background-color);

+    .mdc-list-item__text {
+      align-self: center;
+     display: flex;
+      justify-content: center;
+      flex-direction: column;
+      gap: 2px; /* since flex-gap is not supported currently in safari we might want to set margin only on secondary text */
    }

+    .mdc-list-item__primary-text,
+    .mdc-list-item__secondary-text {
+      display: block;
+     margin: 0px;
+      color: var(--list-item-deselected-color);
+   }

+    .mdc-list-item__primary-text::before,
+    .mdc-list-item__primary-text::after {
+      display: none;
+   }

+    .mdc-list-item__secondary-text::before,
+   .mdc-list-item__secondary-text::after {
+      display: none;
+    }

    &:hover {
      background-color: var(--list-item-deselected-background-color-hover);
    }

    &:focus {
      background-color: var(--list-item-deselected-background-color-focus);
    }

    &:active {
      background-color: var(--list-item-deselected-background-color-active);
    }
    &.mdc-list-item--selected,
    &.mdc-list-item--activated {
      background-color: var(--list-item-selected-background-color);
      color: var(--list-item-selected-color);

      .mdc-list-item__primary-text,
      .mdc-list-item__secondary-text {
        color: var(--list-item-selected-color);
      }

      &::before,
      &::after {
        display: none;
      }

      &:hover {
        background-color: var(--list-item-selected-background-color-hover);
      }

      &:focus {
        background-color: var(--list-item-selected-background-color-focus);
      }

      &:active {
        background-color: var(--list-item-selected-background-color-active);
      }
    }
  }
}

```

P1NHE4D: That's an elegant solution. Thank you for your help!

PR #339, opened 2021-04-08
feat(elements|ino-img): add fallback img

### LINKED ISSUE #295: ino-img: Add image fallback when no image available or loadable
According to https://designsystemchecklist.com/category/core-components there should be fallbacks when there’s no image available to load. This can be done with placeholder images or initials.

* In case of an error, an icon is displayed. The icon is an SVG icon and has a fixed maximum size.
* The background is "light gray".
* The SVG icon can be replaced by the consumer with its own SVG icon.

image
image

[image-icons.zip](https://github.com/inovex/elements/files/6050344/image-icons.zip)

P1NHE4D:
✨ **Features:**
* added fallback image with customizable icon to `ino-img` component

🎟️ **Closes #295**


packages/elements/src/components/ino-img/ino-img.scss
```diff
@@ -30,4 +29,34 @@ ino-img {
   &.mdc-image-list__item {
     display: list-item;
   }
+
+  img:before {
```
JCofman, line 33: Hi @P1NHE4D is there a reason why we need the pseudo before element for the image?

packages/elements/src/components/ino-img/ino-img.tsx
```diff
@@ -138,6 +144,10 @@ export class Image implements ComponentInterface {
     this.composedRatioHeight = `${ratio}%`;
   }
 
+  private handleError() {
+    this.el.querySelector('.ino-img__fallback-icon').setAttribute("style", "display: block");
```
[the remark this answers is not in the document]
P1NHE4D: The `div` element with the `ino-img__fallback-icon` class is hidden by default as the icon should only be displayed if the image failed to load.

packages/elements/src/components/ino-img/ino-img.scss
P1NHE4D: The pseudo element is only displayed if the image fails to load. As far as I remember, I only added it to hide the default icon and alt text in case of an error while loading the image.

packages/elements/src/assets/ino-icon/image_not_available.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
P1NHE4D: Done ✅ Interesting tool though - I've never seen this before.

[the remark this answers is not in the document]
JCofman: Awesome :) we may could also add a task to automate the process at build time using https://github.com/svg/svgo

packages/elements/src/components/ino-img/ino-img.scss
```diff
@@ -30,4 +29,34 @@ ino-img {
...
+
+    ino-icon {
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      width: max(20px, 10%);
+      height: max(20px, 10%);
+      --icon-width: 100%;
+      --icon-height: 100%;
```
janivo, line 59: The fallback icon is not fully centered because of those two lines. I think you can safely delete them.

packages/elements/src/components/ino-img/ino-img.tsx
```diff
@@ -122,6 +127,7 @@ export class Image implements ComponentInterface {
     if (!this.inoImgListItem) {
       this.el.querySelector('div').style.paddingTop = this.composedRatioHeight;
     }
+    this.el.querySelector('img').onerror = () => this.handleError();
```
janivo, line 130: Could you replace the `querySelector` with an image ref?

e.g.:
```jsx
<img
  // ...
  ref={(ref) => this.imageEl = ref}
/>
```

janivo, line 130:
```js
this.imageEl.onerror = this.handleError;
```
should work

packages/elements/src/components/ino-img/ino-img.tsx
janivo, line 148: Use `imageEl` here

janivo, line 148: Instead of using inline styles, we should add a class to the element which displays the fallback image (e.g. `ino-img__fallback-icon--show`).

janivo (changes requested):

packages/elements/src/components/ino-img/ino-img.tsx
P1NHE4D: ✅

packages/elements/src/components/ino-img/ino-img.tsx
P1NHE4D: ✅

janivo (approved):

PR #357, opened 2021-04-16
fix(elements|ino-datepicker): only validate if value is set

### LINKED ISSUE #356: ino-datepicker: min/max dates set validation state if no value is set
### Describe the bug

Changes to minimum/maximum Date trigger the validation logic even if no value is set.

### To Reproduce

1. Go to https://elements.inovex.de/dist/latest/storybook/?path=/story/input-ino-datepicker--default-usage
2. Delete `value` & `ino-default-date` in `Date Config`
3. Change minimum Date in html via browser developer tools
4. See invalid state 

grafik

Expected behavior:
Only validate if Value is set

Sl1nd:
Closes #356 

## Proposed Changes

- Trigger validation logic only if value is truthy

### DIFF
--- a/packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
+++ b/packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
@@ -55,7 +55,7 @@ export class Datepicker implements ComponentInterface {
   @Watch('required')
   requiredChanged(required: boolean) {
     this.validator.isRequired = required;
-    this.isValid = this.validator.validate(this.value);
+    this.validate();
   }
 
   /**
@@ -72,13 +72,9 @@ export class Datepicker implements ComponentInterface {
 
   @Watch('value')
   valueChanged(value?: string) {
-    this.isValid = this.validator.validate(value);
+    this.validate(value);
 
-    if (this.disabled) {
-      return;
-    }
-
-    if (this.isValid) {
+    if (!this.disabled && this.isValid) {
       this.flatpickr?.setDate(value, true);
     }
   }
@@ -91,7 +87,7 @@ export class Datepicker implements ComponentInterface {
   @Watch('min')
   minChanged(min: string) {
     this.validator.minDate = min;
-    this.isValid = this.validator.validate(this.value);
+    this.validate();
     this.flatpickr?.set('minDate', min);
   }
 
@@ -103,7 +99,7 @@ export class Datepicker implements ComponentInterface {
   @Watch('max')
   maxChanged(max: string) {
     this.validator.maxDate = max;
-    this.isValid = this.validator.validate(this.value);
+    this.validate();
     this.flatpickr?.set('maxDate', max);
   }
 
@@ -139,15 +135,17 @@ export class Datepicker implements ComponentInterface {
   @Prop() inoRange?: boolean;
 
   @Watch('inoRange')
-  inoRangeChanged() {
-    if (!this.disabled) {
-      this.create();
-    }
+  inoRangeChanged(val: boolean) {
+    this.create();
+    this.validator.isRanged = val;
+    this.validate();
   }
 
   @Watch('disabled')
   disabledChanged(newValue: boolean) {
-    newValue ? this.dispose() : this.create();
+    this.create();
+    this.validator.isDisabled = newValue;
+    this.validate();
   }
 
   /**
@@ -160,7 +158,7 @@ export class Datepicker implements ComponentInterface {
   @Watch('inoDateFormat')
   inoDateFormatChanged(dateFormat: string) {
     this.validator.dateFormat = dateFormat;
-    this.isValid = this.validator.validate(this.value);
+    this.validate();
     this.flatpickr?.set('dateFormat', dateFormat);
   }
 
@@ -214,7 +212,7 @@ export class Datepicker implements ComponentInterface {
 
   @Watch('inoType')
   inoTypeChanged() {
-    if (!this.disabled) this.create();
+    this.create();
   }
 
   /**
@@ -270,13 +268,15 @@ export class Datepicker implements ComponentInterface {
   }
 
   componentDidLoad() {
-    if (!this.disabled) {
-      this.create();
-    }
+    this.create();
   }
 
   disconnectedCallback() {
     this.dispose();
+  }
+
+  private validate(value: string = this.value) {
+    this.isValid = this.validator.validate(value);
   }
 
   private create() {
--- a/packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
+++ b/packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
@@ -160,6 +160,7 @@ export class Datepicker implements ComponentInterface {
   @Watch('inoDateFormat')
   inoDateFormatChanged(dateFormat: string) {
     this.validator.dateFormat = dateFormat;
+    this.isValid = this.validator.validate(this.value);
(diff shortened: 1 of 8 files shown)


packages/elements/src/components/ino-datepicker/validation.ts
```diff
@@ -3,58 +3,191 @@ import flatpickr from 'flatpickr';
...
+    this.minDate = minDate;
+    this.maxDate = maxDate;
+  }
+
+  /**
+   * Validates the date in terms of format, require-state, disabled-state minimum date and maximum date.
+   * @param value The value to be validated
+   * @return true if the date is valid, false if the date is invalid
+   */
+  public validate = (value?: string): boolean => {
```
Sl1nd, line 44: Could we use isValid? IMO this would be more explicit considering the return values

packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -180,6 +159,7 @@ export class Datepicker implements ComponentInterface {
 
   @Watch('inoDateFormat')
   inoDateFormatChanged(dateFormat: string) {
+    this.validator.dateFormat = dateFormat;
```
Sl1nd, line 162: Should we also validate if dateformat changes? Maybe there is a valid usecase for dynamic format changes

packages/elements/src/components/ino-datepicker/validation.ts
```diff
@@ -3,58 +3,191 @@ import flatpickr from 'flatpickr';
...
+  maxDate?: string;
 };
 
-export const validateMax = (
-  value: string,
-  format: string,
-  maxDateStr?: string
-): boolean => {
-  if (!maxDateStr) {
+export class DateValidator {
```
Sl1nd, line 15: Maybe Validator as class name is more universal. `Date` only relates to the type Dates. e.g. datetime, month as well as time are not covered

packages/elements/src/components/ino-datepicker/validation.ts
```diff
@@ -3,58 +3,191 @@ import flatpickr from 'flatpickr';
...
+  public validate = (value?: string): boolean => {
+    if (this._isDisabled) {
+      return true;
+    }
+
+    if (!value && this._isRequired) {
+      return false;
+    }
+
+    if (!this.validateFormat(value)) {
```
Sl1nd, line 53: validateFormat and validateMax/validateMin has to be performed in parallel. Otherwise values outter scope  would produce valid date input

Sl1nd: e.g.

if(!value) return this._isRequired;

if(value)   return  validateMin&& validateMax && validateFormat;

see former implementation

packages/elements/src/components/ino-datepicker/validation.ts
janivo: IMHO `isValid` would describe a variable rather than a function that calculates a validity status

packages/elements/src/components/ino-datepicker/validation.ts
janivo: Do you have a concrete example where the current implementation would fail?
I could write it the way you described but I think this way the readability is better.

Sl1nd: Sorry my bad. I realized that the logic work as expected. The Issue that occured is due to parsing date in min and max setter. If the `minDate` is initialized asynchronous in any framework the value may be initially 'undefined' and the datepicker errors out. Maybe we should adjust the setter logic to only parse Values if value is not undefined

packages/elements/src/components/ino-datepicker/validator.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  set minDate(value: string | undefined) {
+    if (!value) {
+      this._minDate = undefined;
+      return;
+    }
+
+    try {
+      this._minDate = this.parseDate(value);
```
Sl1nd, line 159: The issue described below produces unnecessary console errors if `minDates` are initialized asynchronously. Maybe the `minDate`as well as `maxDate` only be parsed if it is not falsy (null|undefined|'"') otherwise it is set to undefined

Sl1nd:
Currently the min/max changes are already validated if no value is present (storybook: value: null, default-date: null & required true -> change min date in Dev-Tools DOM).

janivo:
> Currently the min/max changes are already validated if no value is present (storybook: value: null, default-date: null & required true -> change min date in Dev-Tools DOM).

That's expected because in that case, `null` is just handled as a string. If you want to provide an empty value to the `ino-datepicker` you should use `''` instead of `null` like this:
```jsx
<InoDatepicker ... value={this.myDate || ''} />
```

Sl1nd:
> > Currently the min/max changes are already validated if no value is present (storybook: value: null, default-date: null & required true -> change min date in Dev-Tools DOM).
> 
> That's expected because in that case, `null` is just handled as a string. If you want to provide an empty value to the `ino-datepicker` you should use `''` instead of `null` like this:
> 
> ```js
> <InoDatepicker ... value={this.myDate || ''} />
> ```
This is not an issue of handling the Datepicker. In Storybook you may observe this behavior if you adjust the Date Config  Knob (Delete initial `value`, `ino-default-date` as well as setting required in Standard Knob

Sl1nd:
Additionally i observe some error logs in React context using the datepicker in form on initalization:
grafik

grafik
grafik
grafik
grafik

janivo:
> Additionally i observe some error logs in React context using the datepicker in form on initalization:
> grafik
> 
> grafik
> grafik
> grafik
> grafik

I see. These errors might be related to the initialization of the validator which happens after the component rendered in the DOM (concrete in `componentDidLoad`). I moved the initialization of the validator to the `componentWillLoad` lifecycle. Can you check again @Sl1nd?

Sl1nd:
> > Additionally i observe some error logs in React context using the datepicker in form on initalization:
> > grafik
> > grafik
> > grafik
> > grafik
> > grafik
> 
> I see. These errors might be related to the initialization of the validator which happens after the component rendered in the DOM (concrete in `componentDidLoad`). I moved the initialization of the validator to the `componentWillLoad` lifecycle. Can you check again @Sl1nd?

Still present. I found a solution as i realized that the errors occure in all Watchers on setting the state for the validator. So checking if validator is present before setting state did fix the errors: e.g.

Sl1nd:
> > > Currently the min/max changes are already validated if no value is present (storybook: value: null, default-date: null & required true -> change min date in Dev-Tools DOM).
> > 
> > 
> > That's expected because in that case, `null` is just handled as a string. If you want to provide an empty value to the `ino-datepicker` you should use `''` instead of `null` like this:
> > ```js
> > <InoDatepicker ... value={this.myDate || ''} />
> > ```
> 
> I think this is not an issue of handling the Datepicker. In Storybook you may observe this behavior if you adjust the `Date Config` Knob (Delete initial `value`, `ino-default-date` as well as setting` required` in `Standard` Knob
> Refer to issue description #356

Still the minimum and maximum validation is triggered if no value is set. We need to adjust the  min/max validation methods
e.g.:

```
private validateMinSingle = (value: string): boolean => {
    if(!value) return true;
    const parsedDate: Date = this.parseDate(value);
    return (
      isEqual(parsedDate, this._minDate) || isAfter(parsedDate, this._minDate)
    );
  };```

Sl1nd:
@janivo  Works as expected. From my perspective this PR may be merged.

janivo (approved):

PR #370, opened 2021-05-10
style(elements|ino-list): add css variable for border radius

### LINKED ISSUE #364: ino-list: border-radius doesn't affect hover/focus styles
### Describe the bug
When giving the `ino-list` a custom `border-radius`, the `border-radius` is applied as expected but the hover/focus styles of the first/last `list-item` are not adjusted to the new `border-radius`.

### To Reproduce

1. Go to https://elements.inovex.de/dist/v3.0.1/storybook/?path=/story/structure-ino-list--list-item
2. Scroll down to e.g. 'Graphic and Meta'
4. Give the `ino-list` a `border-radius` (e.g. `border-radius: 26px`)
5. Hover/focus the first or last `ino-list-item`
6. See overlapping corners

#### How it looks
image

#### Expected
image

### Possible Solution
Give the first/last `ino-list-item` in an `ino-list` the following styles:
```scss
  border-(top|bottom)-(left|right)-radius: 13px; // this should be half the given border radius of `ino-list-item`
  overflow: hidden;
```

janivo:
Closes #364 

## Proposed Changes

- Add CSS-Var `--border-radius` to give the container of the `ino-list` a border-radius
- Also affects the first and last `ino-(list|control)-item` in the list

### DIFF
--- a/packages/elements/src/components/ino-list/ino-list.scss
+++ b/packages/elements/src/components/ino-list/ino-list.scss
@@ -4,7 +4,7 @@
 @include list.core-styles;
 
 /**
-  * @prop --border-radius: Border radius
+  * @prop --border-radius: Border radius of the list container
  */
 ino-list {
   --border-radius: (var(--border-radius), 0);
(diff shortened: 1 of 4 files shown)


packages/elements/src/components/ino-list/ino-list.scss
```diff
@@ -1,11 +1,37 @@
 @use 'base/mdc-customize';
 @use '@material/list';
+
 @include list.core-styles;
 
+/**
+  * @prop --border-radius: Border radius of the list container
+ */
 ino-list {
+  --border-radius: (var(--border-radius), 0);
```
P1NHE4D, line 10: Shouldn't it conform to the naming convention used in other components, i.e., `--list-item-border-radius: (var(--ino-list-item-border-radius), 0);`?

P1NHE4D (commented):
I couldn't find the CSS variable in the corresponding story. Is it supposed to be like that?

packages/elements/src/components/ino-list/ino-list.scss
janivo: Yes, you are right! I changed it. Could you review it again?

P1NHE4D (approved):

PR #397, opened 2021-07-02
chore(elements): expose custom types

pfecht:
Exposes our custom types and the component types.

## Proposed Changes

- ...

## Things to check


janivo:
Does this change the way the consumer imports our types?

pfecht:
No, no changes for the consumer. Only more flexibility :)

janivo (approved):

PR #408, opened 2021-08-02
feat(elements|ino-tab-bar): add autoFocus property

### LINKED ISSUE #400: ino-tab-bar: configure autofocus on activate
### Feature Request
In some cases it is necessary to control the autofocus mechanism of the tab bar, e.g. to keep the focus on the nav-drawer element.

Provide the interface of MDCTabBar `focusOnActivate` for ino-tab-bar.
https://material.io/develop/web/supporting/tab-bar

### Additional context

If the ino-tab-bar is used in comparision to the nav-drawer the focus is immediatly set to the tab-bar after navigation. In order prevent a11y issues the focus shift should not be forced by component default.

BenPag:
Closes #400

## Proposed Changes

- add `autoFocus` property to ino-tab-bar component

### DIFF
--- a/packages/elements/src/components/ino-tab-bar/ino-tab-bar.e2e.ts
+++ b/packages/elements/src/components/ino-tab-bar/ino-tab-bar.e2e.ts
@@ -19,6 +19,17 @@ describe('InoTabBar', () => {
       const activeTab = await tabBar.getAttribute('active-tab');
       expect(activeTab).toBe('3');
     });
+
+    it('should render with the autofocus property set correctly', async () => {
+      const page = await setupPageWithContent(INO_TAB_BAR);
+      const tabBar = await page.find(TAB_BAR_SELECTOR);
+
+      await tabBar.setAttribute('autoFocus', true);
+      await page.waitForChanges();
+
+      const autofocus = await tabBar.getAttribute('autoFocus');
+      expect(autofocus).toBe('true');
+    });
   });
 
   describe('Events', () => {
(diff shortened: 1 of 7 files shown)


packages/elements/src/components/ino-tab-bar/ino-tab-bar.e2e.ts
```diff
@@ -19,6 +19,17 @@ describe('InoTabBar', () => {
       const activeTab = await tabBar.getAttribute('active-tab');
       expect(activeTab).toBe('3');
     });
+
+    it('should render with the autofocus property set correctly', async () => {
```
janivo, line 23: This test only checks whether stencil sets the property correctly. I would suggest querying the focused element on that page and checking if it's the intended one. We also got a utility function you can use at `packages/elements/src/util/test-utils.ts`.

Here is an example: https://github.com/inovex/elements/blob/5167724116b44e8d384852b56a42177878dab59c/packages/elements/src/components/ino-input/ino-input.e2e.ts#L11-L27

packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
```diff
@@ -37,6 +37,11 @@ export class TabBar implements ComponentInterface {
     }
   }
 
+  /**
+   * Autofocus of tab on activation.
+   */
+  @Prop() autoFocus?: boolean = false;
```
janivo, line 43: Could you also add that new property to the story of the `ino-tab-bar`?

janivo (changes requested):

packages/storybook/src/stories/ino-tab-bar/ino-tab-bar.stories.ts
```diff
@@ -46,6 +47,12 @@ Playground.argTypes = {
     },
     options: [0, 1, 2, 3],
   },
+  autoFocus: {
```
janivo, line 50: We don't have to explicitly declare the control for primitive types like `boolean`, `number`, and `string`. This can be safely removed.

janivo (changes requested):
Nice 👍🏼  
Just one last thing and we are good to go

janivo (approved):

Issue #348, opened 2021-04-13
ino-select: required validation on form submit is not working


AlessaRad:
### Describe the bug

An empty ino-select that is marked as required is not validated on form submit

### To Reproduce

1. Go to https://elements.inovex.de/dist/latest/storybook/?path=/story/input-ino-select--forms
2. Try to submit the form without selecting a value in the ino-select
3. The form is submitted, indicated by the alert.

### Expected behavior

The form should not submit and the ino-select should be marked as invalid (red).

### Screenshots

![image](https://user-images.githubusercontent.com/76041234/114548810-8b223380-9c60-11eb-8ffc-50dfe7c24a97.png)

### Possible Solution

I assume, that the validation is missing, because the native HTML validation on form submit only validates native input fields. The ino-select only has a hidden input for synchronizing the value. Hidden inputs are ignored on form submit.

A possible solution is to use a non-hidden input and hide it via the css display property (none).
Furthermore a Listener on the the 'invalid' event that is thrown by checkValidity() which is called on the native input field when a form is submitted, can invalidate the mdcSelect instance.
This way works, but it should be evaluated further.

ino-select.tsx:
`...<Host name={this.name}>
        <div class={classSelect}>
          {this.required && <input
            ref={(el) => (this.nativeInputElement = el)}
            required={this.required}
          >
          </input>}
          <div class="mdc-select__anchor" aria-required={this.required}>...`

`... componentDidLoad() {
this.nativeInputElement.addEventListener('invalid', () => {
      this.mdcSelectInstance.valid = false;
    });
}`

ino-select.scss:
`input {
    display: none;
  }
`

JCofman:
Hi @AlessaRad,

awesome thanks for your write-up :). As I looked into the issue I would also suggest changing the hidden `<input>` to a hidden native `<select>` element which is a more appropriate DOM element for this component.

AlessaRad:
Hi @JCofman, I think the select element would be a valid alternative, if the value synchronisation still works. In the MDC select documentation the suggest the hidden input element: https://material.io/develop/web/components/input-controls/select-menus see section "Select with hidden input (for HTML forms)"

PR #416, opened 2021-08-17
fix(elements|ino-select): hide input used for form validation

### LINKED ISSUE #414: ino-select: Native select element should be hidden
### Bug Report

A native select element is displayed inside our `ino-select` element. See below.

image

The error does not appear in the storybook because of [these styles](https://github.com/inovex/elements/blob/666fd2132f42e91dd0a5b81a37c4d7a0d039a5b5/packages/storybook/config/global.scss#L16-L27) which were added in the same PR by mistake.

### Is this a regression?

Did this behavior use to work in the previous version?

- [ ] I don't know
- [x] Yes, the previous version in which this bug was not present was: `v4.0.0`
- [ ] No

### Expected behavior

We use a hidden native `<select>` to make the `ino-select` work with the native `<form>`.
- the native `<select>` should not be shown and additionally
- the state of the `<ino-select>` (options and selected value) should be synced to the hidden native `<select>`.

### Possible Solution

- Move the above mentioned styles from `packages/storybook/config/global.scss` to `packages/elements/src/components/ino-select/ino-select.scss`
- Test if a simple `display: hidden` suffices
- A test should be added** to confirm the correct form handling (e.g. form should not be submittable if the select is required).

janivo:
Closes #414. Basically reverts https://github.com/inovex/elements/pull/355 and fixes https://github.com/inovex/elements/issues/348 properly.

## Proposed Changes

- Revert to using a hidden input instead of select
- Move hidden input to prevent collision with mdc select internal logic
- Add tests to confirm correct form submission handling
- Add a story to test form submission

### DIFF
--- a/packages/storybook/src/stories/ino-select/ino-select.stories.ts
+++ b/packages/storybook/src/stories/ino-select/ino-select.stories.ts
@@ -134,7 +134,7 @@ export const Form = () => html`
   <form>
     <p>Form should not be submittable if no value is selected</p>
     <ino-select required> ${optionsTemplate}</ino-select>
-    <button disabled type="submit">Submit</button>
+    <button type="submit">Submit</button>
   </form>
 `;
 
--- a/packages/storybook/src/stories/ino-select/ino-select.stories.ts
+++ b/packages/storybook/src/stories/ino-select/ino-select.stories.ts
@@ -134,7 +134,7 @@ export const Form = () => html`
   <form>
     <p>Form should submit if no value is selected</p>
     <ino-select required> ${optionsTemplate}</ino-select>
-    <button type="submit">Submit</button>
+    <ino-button type="submit">Submit</ino-button>
   </form>
 `;
 
--- a/packages/storybook/src/stories/ino-select/ino-select.stories.ts
+++ b/packages/storybook/src/stories/ino-select/ino-select.stories.ts
@@ -2,8 +2,15 @@ import { Components } from '@inovex.de/elements';
 import { useEffect } from '@storybook/client-api';
 import { Story } from '@storybook/web-components';
 import { html } from 'lit-html';
-import { defaultDecorator } from '../utils';
+import { defaultDecorator, showSnackbar } from '../utils';
 import './ino-select.scss';
+
+const handleFormSubmission = (e) => {
+  e.preventDefault();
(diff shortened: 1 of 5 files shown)


packages/storybook/src/stories/ino-select/ino-select.stories.ts
```diff
@@ -119,6 +130,14 @@ export const Required = () => html`
   </ino-select>
 `;
 
+export const Form = () => html`
+  <form>
+    <p>Form should submit if no value is selected</p>
+    <ino-select required> ${optionsTemplate}</ino-select>
+    <button type="submit">Submit</button>
```
BenPag, line 137: The button is not styled so it looks unsuitable.

packages/storybook/src/stories/ino-select/ino-select.stories.ts
```diff
@@ -2,9 +2,16 @@ import { Components } from '@inovex.de/elements';
 import { useEffect } from '@storybook/client-api';
 import { Story } from '@storybook/web-components';
 import { html } from 'lit-html';
-import { defaultDecorator } from '../utils';
+import { defaultDecorator, showSnackbar } from '../utils';
 import './ino-select.scss';
 
+const handleFormSubmission = (e) => {
```
BenPag, line 8: Form submit not workings as expeckted. When I click on the button, an empty snackbar opens (In both cases - empty submit & filled submit)

BenPag (changes requested):
When I start my local storybook I see some design issues on the ino-select story page.

image

packages/storybook/src/stories/ino-select/ino-select.stories.ts
janivo: Huh, I can't really reproduce the issues you are seeing. Did you start the storybook with `yarn dev` inside the storybook package by any chance? In that case, make sure that you've run `yarn build` inside the elements package before. You can also just use the `yarn start` script on root level which combines these two steps.

BenPag: I retried it with your instructions and now it looks good and everything works fine. I think something was not build correctly.

BenPag (approved):

PR #417, opened 2021-08-17
feat(elements|ino-select): add custom error property and fix error on…

pfecht:
… slot changes

Closes #...

## Proposed Changes

- ...

## Things to check


janivo:
To-Do's:
- [ ] merge with #416 
- [ ] adjust `<ino-autocomplete>`
	- [ ] to work with new options component
	- [ ] to work with mutation event

janivo:
This PR handles many things at once. That's why I decided to split it into smaller PR's.
- Add error state: #424 
- ...

PR #422, opened 2021-09-06
fix(storybook): scope all story styles

janivo:
Closes /

## Proposed Changes

- rename function `defaultDecorator` to `decorateStoryWithClass` for clarity
- correctly decorate all stories (the function was missing or misused sometimes)
- scope all story styles
- add the `Meta` type to all story exports to enable type checking

### DIFF
--- a/packages/storybook/src/stories/ino-option-group/ino-option-group.stories.ts
+++ b/packages/storybook/src/stories/ino-option-group/ino-option-group.stories.ts
@@ -1,12 +1,13 @@
+import { Meta } from '@storybook/web-components';
 import { html, TemplateResult } from 'lit-html';
-import { defaultDecorator } from '../utils';
+import { decorateStoryWithClass } from '../utils';
 import './ino-option-group.scss';
 
 export default {
   title: 'Input/ino-option-group',
   component: 'ino-option-group',
-  decorators: [defaultDecorator],
-};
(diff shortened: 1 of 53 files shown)


packages/storybook/src/stories/ino-option-group/ino-option-group.stories.ts
```diff
@@ -1,12 +1,13 @@
...
-import { defaultDecorator } from '../utils';
+import { decorateStoryWithClass } from '../utils';
 import './ino-option-group.scss';
 
 export default {
   title: 'Input/ino-option-group',
   component: 'ino-option-group',
-  decorators: [defaultDecorator],
-};
+  decorators: [(story) => decorateStoryWithClass(story, 'story-option-group')],
```
BenPag, line 9: The does css class `story-option-group-customizable` wrapped in `story-option-group` too? Currently only `ino-select` is wrapped.

BenPag (changes requested):

packages/storybook/src/stories/ino-option-group/ino-option-group.stories.ts
janivo: Whops, my bad. Should be fixed now.

BenPag (approved):

PR #425, opened 2021-09-13
(elements|ino-dialog): enable extend styling support

BenPag:
## Proposed Changes

- add additional CSS custom properties to extend styling features
  -  `--ino-dialog-border-radius`
  -  `--ino-dialog-box-shadow`
  - `--ino-dialog-border`
  -  `--ino-dialog-overlay-background-color`


packages/elements/src/components/ino-dialog/ino-dialog.scss
```diff
@@ -16,9 +20,24 @@
   --dialog-header-height: var(--ino-dialog-header-height, 20%);
   --dialog-content-height: var(--ino-dialog-content-height, 60%);
   --dialog-footer-height: var(--ino-dialog-footer-height, 20%);
+  --dialog-border-radius: var(--ino-dialog-border-radius, unset);
```
janivo, line 23: These new properties should be added to the comment above (`@prop --dialog-header-height: Height of the header`, ...) so we can generate the table in our readme.

janivo (changes requested):

packages/elements/src/components/ino-dialog/ino-dialog.scss
BenPag: Okay, thank you. I changed this now.

Otherwise, everything is fine?

packages/elements/src/components/ino-dialog/ino-dialog.scss
```diff
@@ -9,16 +13,35 @@
    * @prop --ino-dialog-header-height: height of the header section
    * @prop --ino-dialog-content-height: height of the content section
    * @prop --ino-dialog-footer-height: height of the footer section
+   * @prop --ino-dialog-border-radius: border-radius of the dialog
+   * @prop --ino-dialog-border: border shorthand of the dialog
+   * @prop --ino-dialog-box-shadow: box-shadow of the dialog
+   * @prop --ino-dialog-overlay-background-color: background color of the overlay
    */
   --dialog-background-color: var(--ino-dialog-background, #fff);
```
janivo, line 21: `--ino-dialog-background-color` does not get picked up (because of `--dialog-background-color: var(--ino-dialog-background, #fff);`) -> `...-color` is missing

packages/elements/src/components/ino-dialog/ino-dialog.scss
```diff
@@ -9,16 +13,35 @@
...
   --dialog-width: var(--ino-dialog-width, 300px);
```
janivo, line 22: What do you think about using `auto` as defaults for `ino-dialog-height` and `ino-dialog-width` so the box is automatically as big as it needs to be?

janivo (changes requested):
In general, I would suggest to use `packages/elements/src/components/ino-icon-button/ino-icon-button.scss` as inspiration. We should prefer SCSS-Vars over CSS-Vars internally unless we need to do some more complex stuff (calc, ...).

packages/elements/src/components/ino-dialog/ino-dialog.scss
BenPag: For me it´s fine. 👍

BenPag:
I agree with your suggestion and so I switched over to SCSS-Vars.

PR #427, opened 2021-09-13
feat(elements|ino-snackbar): add new property `verticalAlignment`

### LINKED ISSUE #426: ino-snackbar: Add vertical alignment options
### Feature Request

Currently, it is only possible to customize the horizontal alignment of the `<ino-snackbar>`. In some cases, it might be necessary to customize the vertical alignment of a `<ino-snackbar>` too (e.g.: to show it at the top of the screen).

Add a new property `verticalAlignment` which takes either `top` or `bottom` as its input.

For our new design `top` should be the default.

janivo:
Closes #426 

## Proposed Changes

- add new property `verticalAlignment`
- rename `alignment` to `horizontalAlignment` **(BREAKING)**

### DIFF
--- a/packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
+++ b/packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
@@ -36,9 +36,14 @@ export class Snackbar implements ComponentInterface {
   @Prop() actionText?: string = '';
 
   /**
+   * Controls if Snackbar will appear at the top or at the bottom of the screen
+   */
+  @Prop() verticalAlignment?: 'top' | 'bottom' = 'top';
+
+  /**
    * Controls if Snackbar is centered or left-aligned or right-aligned.
    */
-  @Prop() alignment: 'left' | 'right' | 'center' = 'center';
+  @Prop() horizontalAlignment: 'left' | 'right' | 'center' = 'right';
 
   /**
    * Changes the snackbar type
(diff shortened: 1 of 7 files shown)


packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
```diff
@@ -35,10 +35,15 @@ export class Snackbar implements ComponentInterface {
...
+  /**
+   * Controls if Snackbar will appear at the top or at the bottom of the screen
+   */
+  @Prop() verticalAlignment?: 'top' | 'bottom' = 'top';
+
   /**
    * Controls if Snackbar is centered or left-aligned or right-aligned.
    */
-  @Prop() alignment: 'left' | 'right' | 'center' = 'center';
+  @Prop() horizontalAlignment: 'left' | 'right' | 'center' = 'right';
```
BenPag, line 46: ` 'center' = 'right'` ?

janivo: Yes, according to the designs, `right` should be the new default value.

BenPag: Oh, now I can see it 🙈 I thought that 'right' was assigned to 'center', so never mind 😂

BenPag: All made changes are good.  By testing it via storybook I recordnize that multiply instances were opened on clicking the button. I think the event target are not decoupled. Should I open an issue for that?

BenPag (approved):

Issue #423, opened 2021-09-09
inputs: add form validation functionalities


janivo:
## Feature Request

When using our input elements (`ino-input`, `ino-select`, `ino-checkbox`, ...) I want to be able to use them in a form with the relevant functionalities.

### Required
Add the `required` attribute to all input elements and check if an invalid state is already implemented

- [ ] ino-autocomplete
- [ ] ino-checkbox
- [ ] ino-datepicker
- [x] ino-input
- [ ] ino-input-file
- [ ] ino-radio(-group)
- [ ] ino-range
- [ ] ino-select
- [ ] ino-switch
- [ ] ino-textarea

*open a new issue, if there is no visual indicator for an invalid state* 

### Focus & Blur
For some validation mechanisms, it is necessary to trigger the `blur` & `focus` methods of the input components.

- [ ] ino-autocomplete
- [ ] ino-checkbox
- [ ] ino-datepicker
- [ ] ino-input
- [ ] ino-input-file
- [ ] ino-radio(-group)
- [ ] ino-range
- [ ] ino-select
- [ ] ino-switch
- [ ] ino-textarea

### Validation States & Message
As is the case for the `ino-input`, we should provide properties to mark the component as invalid from the outside, following our controlled components concept. Furthermore, it is necessary to provide a way to show custom error messages as well.

- [ ] ino-autocomplete
- [ ] ino-checkbox
- [ ] ino-datepicker
- [ ] ino-input
- [ ] ino-input-file
- [ ] ino-radio(-group)
- [ ] ino-range
- [ ] ino-select
- [ ] ino-switch
- [ ] ino-textarea

### Summary
In the end, every input component should have the following methods/properties (names are WIP):
- `required`
- `setFocus()` 
- `setBlur()`
- `error`
- `helper` (maybe rename to `helperText`?)
- `helperPersistent`
- `helperValidation`
- `showLabelHint`

*for an explanation of each property, have a look at the [API of the `ino-input` element](https://elements.inovex.de/dist/latest/storybook/?path=/docs/input-ino-input)*

janivo:
#306 should be used as a starting point for the `Focus & Blur` functionalities

janivo:
This feature might have a big impact on this story: #1232

PR #429, opened 2021-09-16
feat(elements|ino-datepicker): provide error property

BenPag:
Relates to #423

## Proposed Changes

- add property `error` to enable error state programmatically


packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -244,6 +244,26 @@ export class Datepicker implements ComponentInterface {
     this.flatpickr.toggle();
   }
 
+  /**
+   * Displays the datepicker as invalid if set to true.
+   * If the property is not set or set to false,
+   * the validation is handled by the default validation.
```
pfecht, line 250: Shouldn't the error prop have three states:
* `null|undefined` => use native validation
* `true` => has error
* `false` => has no error

If I integrate the component and manually set the error=false, I wouldn't suggest that the component still natively validates.

This may be something we should unify across all components

pfecht (approved):

PR #431, opened 2021-09-20
feat(elements|ino-snackbar): rework snackbar

janivo:
## Proposed Changes

- Types
	- remove `primary` and `warning` ‼️
 	- add `info` and `success`
	- set `info` as default
	- update design
- Alignment
	- remove `alignment` property ‼️
	- add css vars to make position customizable (`ino-snackbar-(top|right|bottom|left)`) (can be useful to implement stacking)
	- set top-right as the default position
- Design
	- add new designs for `info` and `success`
	- update action-button, shapes and layout according to the design
	- move the close-Button to the top-right edge of the notification window


packages/elements/src/components/ino-snackbar/ino-snackbar.scss
```diff
@@ -4,89 +4,125 @@
...
+    box-shadow-color: #2d5dff29,
+  ),
+  error: (
+    color: #7df481,
+    box-shadow-color: #7df48129,
+  ),
+  success: (
+    color: #f29191,
+    box-shadow-color: #f2919129,
+  ),
```
BenPag, line 21: I think the colors should be swaped. `success` is green and `error` is red.

packages/elements/src/components/ino-snackbar/ino-snackbar.scss
```diff
@@ -4,89 +4,125 @@
...
+);
```
BenPag, line 22: These colors are the new elements colors I guess. I think it is better to use the colors defined in `theme.color` or at least mention it in the issue that hanfles the implementation of the new color pallet.

BenPag (changes requested):
Great Job! Looks very nice 👍

packages/elements/src/components/ino-snackbar/ino-snackbar.scss
janivo: As we discussed earlier, I switched to the old color pallette for now.

janivo:
blocked: waiting for the new icons

PR #432, opened 2021-09-20
refactor: load font `lato` from inovex host

### LINKED ISSUE #160: Refactor fonts
As a consumer, I would like to have the choice of my preferred font. Currently, the use of Lato font is deeply connected inside the elements and wrappers like Angular.

It would be better to let the consumer decide whether the fonts should be imported or not.

### Aspects of a possible solution

* The default font face for all components can be changed at a single place (css variable?)
* No inovex-elements fonts are imported if the consumer changes the font face.
* No font-dependant code in any angular/react wrapper.

BenPag:
Closes #160 

## Proposed Changes

- provide global css var `--ino-font-family` for custom font
- loading _Lato_ font from static inovex host if necessary

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -15,3 +15,5 @@ proxies-list.ts
 proxies.ts
 utils.ts
 CHANGELOG.md
+
+import-fonts.ts
--- a/.prettierignore
+++ b/.prettierignore
@@ -15,5 +15,3 @@ proxies-list.ts
 proxies.ts
 utils.ts
 CHANGELOG.md
-
-import-fonts.ts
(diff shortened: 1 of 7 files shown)


janivo (approved):
Great job! I've only added some type declarations. If you are fine with it, feel free to merge.

janivo:
@BenPag: could you check if the errors in the `build` step occur on your machine as well? If you also cannot reproduce it, I think we might have a bug with the caching in the `install` step.

janivo (approved):

BenPag:
I'm fine with it but can´t merge it. I'm not authorized. 😄

Issue #403, opened 2021-07-30
Replace text and label properties with slots


silentHoo:
### Refactor Request

- [x] ino-form-row: `label` -> default slot?
   - default slot is currently in use, no changes useful
- [x] ino-chip: `label` -> default slot? (#559)
- [x] ino-header: `text` -> default slot? (#543)
- [x] ino-list-item: `text` -> default slot?
   - default slot is currently in use for user content like images
   - named slot `primary` exists
- [x] ino-list-item: `secondary-text` -> named slot?
   - named slot `secondary` exists
- [ ] ino-fab: `label` -> default slot?
- [x] ino-nav-item
   - `text` -> default slot?
   - `secondary-text` -> named slot?
     - Component will be replace by `ino-list-item` in future
- [x] ino-control-item 
  - `text` -> default slot? 
  -  `secondary-text` -> named slot?)
     - depends on `ino-list-item` 
     - is slot delegating possible?
     - can possible replaced by composition example?
- [x] ino-tab: `label` -> default slot? (#544)
- [x] ino-tooltip: `label` -> default slot? (#558)
- [x] ino-snackbar: `message` -> default slot? (#542)
- [x] ino-label: `label` -> default slot?
  - is only a internal component
  - used by `ino-input`, `ino-textarea`, `ino-select`

Please consider valuable choices for each component.

*Acceptance Criteria*:
- Add a deprecation warning
- Allow both solutions: label/text/message and slot
- Add a new issue removing the label/text/message behaviour

BenPag:
Component remvoal ideas:

* ino-form-row
* ino-nav-item (can be replaced by ino-list-item)

PR #470, opened 2021-12-02
refactor(elements|ino-snackbar): use default slot instead of property

BenPag:
**Target branch is not the 'master'** branch !
<hr />

* Relates to #403 
* remove `message` property from ino-snackbar 
* using default slot for message


packages/elements/src/components/ino-snackbar/readme.md
```diff
@@ -121,16 +120,6 @@ Snackbar is displayed when `show` is changed to checked.
 | `hideEl`      | Event that emits as soon as the snackbar hides. Listen to this event to hide or destroy this element. | `CustomEvent<any>` |
 
 
-## CSS Custom Properties
```
janivo, line 124: This was probably removed by accident

janivo (changes requested):

janivo (approved):

PR #483, opened 2022-01-05
feat(elements|ino-dialog): rework the dialog component

### LINKED ISSUE #477: Fullwidth dialogs
### Feature Request
The dialog should have a fullscreen option which floats in from the button of the page for large dialogs (for instance multi step wizards).

Add a `fullwidth` property and restyle dialog.

### Additional context

Component already exists and must be migrated.

pfecht:
BREAKING CHANGE: Remove multi slots and related css variables for header, content, footer slot using default slot. Replace `openChange` event with a `close` event emitting a dialog action string.

Closes #477

## Proposed Changes
* Removes multi slots to use only the default slot
* Adds fullwidth option
* Add dissmissible option for non-critical dialogs.
* Add dialog action
* Rework openChange event to close 
* Reworks storybook of dialog (with #482 in mind)
## Things to check


packages/elements/src/components/ino-dialog/readme.md
```diff
@@ -58,32 +55,48 @@ class MyComponent extends Component {
...
+| `close` | Emits an event upon closing the dialog | `CustomEvent<string>` |
 
 
 ## Slots
 
 | Slot        | Description           |
 | ----------- | --------------------- |
-| `"content"` | content of the dialog |
-| `"footer"`  | footer of the dialog  |
-| `"header"`  | header of the dialog  |
```
janivo, line 86: Why do you prefer a single slot instead of three separate slots?
I know that this is a lot more flexible but wouldn't that speak against our design system concept?

packages/storybook/src/stories/ino-dialog/ino-dialog.stories.ts
```diff
@@ -77,18 +76,65 @@ export const Playground: Story<Components.InoDialog> = (args) => html`
...
+        color-scheme="secondary"
         >Cancel
       </ino-button>
-      <ino-button icon="mail" id="close-dialog-btn">Submit</ino-button>
+      <ino-button icon="mail" data-ino-dialog-action="submit">Submit</ino-button>
     </div>
-  </ino-dialog>
 `;
 
+export const Playground: Story<Components.InoDialog> = (args) => html`
```
janivo, line 91: There seems to be a problem with the playground.
Whenever I try to use the knobs (e.g. `fullwidth="true"`), I only get to see the default story.
I've checked the DOM and it seems like the `<ino-dialog>` elements are duplicated whenever the knobs are used.

How to reproduce:
1. Run the storybook
2. Go to Dialog Story
3. Press the Button `Open Dialog`
4. Close the Dialog
5. Set `fullwidth` to `true` by using the knobs
6. Press the Button `Open Dialog` again
7. See the default dialog instead of a full-width one

janivo (changes requested):

packages/elements/src/components/ino-dialog/readme.md
pfecht: Yeah, as long as we don't have any styling, I would remove the unused slots. They add complexity:
* why is the footer height 20% when it doesn't have any styling?
* why do we need css vars to set the height on a slotted content?

In my opinion, we should use the `ino-dialog` as a generic base component and follow the concept of composition, meaning either:
* for instance `ino-confirmation-dialog` as a child in the dialog providing a special functionality for confirmations,  or a `ino-dialog-wizard` for fullwidth dialogs.
* or style classes for containers in a global.scss being imported.

packages/storybook/src/stories/ino-dialog/ino-dialog.stories.ts
pfecht: Seems to be fixed with `attach-to`, even though I'm not sure why :D Let me know if it works for you as well.

janivo (approved):

PR #486, opened 2022-01-10
feat(elements-angular): remove obsolet event fix and upgrade stencil …

### LINKED ISSUE #487: Angular target not working with `strictTemplate: true`
### Bug Report
Using generated directives is not compatible with  `strictTemplates: true` config in tsconfig. for Angular projects. 

### Minimal Reproduction
```html
<ino-checkbox (checkedChange)="onCheck($event)" name="instruction-dialog-checkbox">
        Tipps
</ino-checkbox>

```

```ts 
 onCheck(event: any) {
    console.log('onCheck ausgeführt');
  }
```

Error: `Property 'subscribe' does not exist on type 'EventEmitter<any>`

### Expected behavior

No error.

### Your Environment

| Software            | Version |
| ------------------- | ------- |
| @inovex.de/elements |         |
| Browser             |         |
| Operating System    |         |

### Possible Solution

see #486

pfecht:
…target generator

Closes  #487 

## Proposed Changes

- ...

## Things to check


packages/elements/stencil.config.ts
```diff
@@ -31,16 +31,9 @@ export const config: Config = {
...
-      directivesUtilsFile:
-        '../elements-angular/elements/src/directives/proxies-utils.ts',
-      directivesArrayFile:
-        '../elements-angular/elements/src/directives/proxies-list.ts',
-    }),
-    angularOutputTargetFix({
-      directivesUtilsFile:
-        '../elements-angular/elements/src/directives/angular-component-lib/utils.ts',
+      directivesProxyFile: `${angularDiretivesPath}/proxies.ts`,
+      directivesUtilsFile: angularDiretivesPath,
```
BenPag, line 35: It seems that we use the same utils [file](https://github.com/ionic-team/stencil-ds-output-targets/blob/master/packages/angular-output-target/angular-component-lib/utils.ts) like the npm package. 
I suggest to clean up our code base and use the default.

JCofman (approved):
LGTM

Issue #262, opened 2021-01-15
Documentation: Restructure Storybook documentation


silentHoo:
We use Storybook to generate the documentation. This approach is now used for over two years. In the meantime Storybook introduced a new documentation format called MDX.

We saw that we need to put much effort in maintaining the documentation and the overall experience from a developer perspective could be much better as it is. Additionally the documentation is a bit unstructured and grown over time. We need to refactor and sort out some old docs and replace it with more structure.

- [ ] Think about how we need to restructure the documentation.
  - [ ] Check the possibilities MDX give us.
  - [ ] Propose the new concept in an GitHub issue and discuss your idea with the team.
  - [ ] Change the docs for one component only to demonstrate the new concept.

A first idea for a good structure could be:
1. One page with general info about the component itself.
2. One page with code examples for all specific frameworks (with a select control to choose the framework).
3. API documentation for events, props and css custom properties.

janivo:
I did some research and came up with the following PR to simplify our stories: https://github.com/inovex/elements/pull/490

All stories should follow the `<ino-checkbox>` example story. 
This way of writing our stories allows us:
- to use typings in our stories which results in fewer errors
- to write less because we can reuse templates
- to include our description in the stories instead of having a README and a story for each component

PR #490, opened 2022-01-14
refactor(storybook|ino-checkbox): use reusable templates for stories

janivo:
## Proposed Changes
- Proposal of a new syntax for easier story maintenance
- uses the [Template syntax of Storybook](https://storybook.js.org/docs/react/writing-stories/introduction#using-args) to create stories

❗  for context, have a look at #262


packages/storybook/src/stories/ino-checkbox/ino-checkbox.stories.ts
```diff
@@ -49,43 +58,17 @@ export const Playground: Story<Components.InoCheckbox> = (args) => html`
...
-      >Checked and disabled
-    </ino-checkbox>
-    <ino-checkbox selection disabled>Unchecked and disabled</ino-checkbox>
-    <ino-checkbox selection indeterminate disabled>
-      Indeterminate Disabled
-    </ino-checkbox>
-  </div>
-`;
+// While checkboxes are mainly used in lists, the selection should be used as a single, independent UI element.
+export const Selection = templateForComponent(Template, {
```
pfecht, line 67: Good to know that this would be possible, but looks too complicated to me. I would keep the inheritance level on 1, meaning
```
export const Selection = templateForComponent(Checked, {indeterminate: true})
```

if possible

packages/storybook/src/stories/ino-checkbox/ino-checkbox.stories.ts
```diff
@@ -49,43 +58,17 @@ export const Playground: Story<Components.InoCheckbox> = (args) => html`
...
 
-Playground.args = {
-  checked: false,
-  disabled: false,
-  indeterminate: false,
-  name: '',
-  selection: false,
-  value: '',
-};
+export const Playground = templateForComponent(Template, {});
```
pfecht, line 61: Should we think the Template idea further? Lets create a "TemplateGenerator", you pass arguments, a component-tag (`ino-checkbox`) and  a typing. Further, it generates a lit html having the args as attribute names and binding its values.

This would might not be suitable for all elements, but at least for extensive  ones like ino-input, ...

packages/storybook/src/stories/ino-checkbox/ino-checkbox.stories.ts
janivo: I agree, it was only meant to show how it's done if we need it e.g. for some complex states

BenPag:
By testing in unicron I figured out, that the Indeterminate-Story not working as I expeced. Afer unchecking the box, I can not check back to the indeterminate state.

packages/elements/src/components/ino-checkbox/readme.md
```diff
@@ -2,121 +2,6 @@
...
-    );
-  }
-}
-```
-
-## Additional Hints
-
-By default, the label is always shown on the right side of the checkbox.
-
-**Indeterminate:** The indeterminate status is set by the user. It indicates that a user is indeterminate without changing the checked state. If a checkbox is unchecked and indeterminate then it will lose the indeterminate state on click and change to checked. [Documentation on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Indeterminate_state_checkboxes)
```
BenPag, line 110: Are these information or desctiptions now not longer important? I can not find any replacements for this.

packages/storybook/src/stories/ino-checkbox/ino-checkbox.stories.ts
```diff
@@ -33,59 +34,51 @@ export default {
...
 
-Playground.args = {
-  checked: false,
-  disabled: false,
-  indeterminate: false,
-  name: '',
-  selection: false,
-  value: '',
-};
+export const Playground = template.generateStory({});
```
pfecht, line 65: Just on idea to make things clearer here: Rename this to `template.generatePlaygroundStory()` (default should be without passing an object)

packages/storybook/src/stories/ino-checkbox/ino-checkbox.stories.ts
```diff
@@ -33,59 +34,51 @@ export default {
...
-  </div>
-  <div class="story-states">
-    <ino-checkbox disabled checked>Checked and Disabled</ino-checkbox>
-    <ino-checkbox disabled>Unchecked and Disabled</ino-checkbox>
-    <ino-checkbox disabled indeterminate
-      >Indeterminate and Disabled
-    </ino-checkbox>
-  </div>
-`;
+export const Checked = template.generateStory({ checked: true }, 'checked');
```
pfecht, line 67: Rename this to `generatePropertyStory(key, value)`, i.e. `generatePropertyStory(checked, true)`.  Would also add some documentation in the function then to clarify that the main feature is that the documentation will be extracted.

packages/storybook/src/stories/ino-checkbox/ino-checkbox.stories.ts
```diff
@@ -33,59 +34,51 @@ export default {
...
-    </ino-checkbox>
-    <ino-checkbox selection disabled>Unchecked and disabled</ino-checkbox>
-    <ino-checkbox selection indeterminate disabled>
-      Indeterminate Disabled
-    </ino-checkbox>
-  </div>
-`;
+export const Selection = template.generateStory(
+  {
+    ...Checked.args, // reuse args of Checked Story
```
pfecht, line 71: Sjhouldnt this be default and not configurable for all stories that share a template generator? Otherwise I would add abstraction to this to hide the actual implementation (`inheritArguments: bool` or so).

packages/storybook/src/stories/template-generator.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Generates a new story based of the provided args.
+   * Additionally, adds a description if the parameter `prop` is used.
+   * The `prop` is used to search for the JSDoc comment of the specified property in the `elements-stencil-docs.json`.
+   *
+   * @param args Properties to assign to the template
+   * @param prop The name of the "main" property the stories is written for
+   */
+  public generateStory(args: Component, prop?: string) {
```
pfecht, line 24: Can we type `prop` better than string to enfore that a property actually exists?

pfecht (changes requested):
Had another look and was pretty strict (we should have this clean if this is the new boilerplate for other components :)) 

Let me know what you think of it.

janivo:
Hey @pfecht,
that's right, we should be really strict with that one as it will be the foundation of all of our other stories. I've applied your feedback. Have a **strict** look at it again!

packages/storybook/src/stories/template-generator.ts
```diff
@@ -0,0 +1,107 @@
...
+    return this.generateTemplate(args, prop);
+  }
+
+  /**
+   * A utility function to inherit the args of one or more stories.
+   *
+   * @param mainStory The main story which arguments have the highest priority
+   * @param storiesToReuse One or more stories which args should be inherited. The last one has the highest priority (still below `mainStory`).
+   */
+  public extendStory(
```
janivo, line 56: Not quite sure about that one. Another option would be to introduce our own class `Story` which extends the storybook type. In there we could provide a method like `inherit`. Something like this:
```ts
class OurStory extends Story<Component> {
 // ...

  inherit(story: Story<Component>) {
        this.args = Object.assign({}, this.args, story.args);
        return this;
  }
}
```

That would allow us to write our stories like this:
`export const Selection = template.generateStory('selection', true).inherit(Checked).inherit(...)`

What do you think @pfecht ?

packages/storybook/src/stories/template-generator.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Generates a new story based of multiple arguments.
+   * Adds a description of the parameter which is provided by `prop`.
+   *
+   * @param args Properties to assign to the template
+   * @param prop The name of the "main" property the stories is written for
+   */
+  public generateStoryWithArgs(args: Component, prop: keyof Component) {
```
pfecht, line 46: private?

packages/storybook/src/stories/template-generator.ts
pfecht: This whole inheritance concept doesn't look that intuitive to me. Is the goal to inherit only the default state of args (because the template is defined be the template builder, isn't?) 

Then I wouldn't support this and provide an additional `defaultValues` field or so to the `generateStory`:

`generateStory('indeterminate', true, {'checked': true})`

What do you think?

janivo: The goal is to minimize copy-pasting of other stories.
Imagine you want to have a Story A and Story B.

Story A is quite complex and has five properties that need to be set. Story B has the same properties but with one exception. You would have to copy the exact same properties to Story B and in the long run that would result in higher maintenance costs.

packages/storybook/src/stories/template-generator.ts
janivo: That's one other way to generate a story. IMO there are three different scenarios:
1. Create a playground story (`generatePlaygroundStory`)
2. Create a story for a single property where no other properties have to be set (`generateStory(...)` e.g.: `generateStory('checked', true)`) 
3. Create a more complex story where you have to set multiple properties at once ( `generateStoryWithArgs`), e.g.: 
```ts
generateStoryWithArgs(
	{ 
		checked: true,
		selection: true,
		disabled: true,
	},
	'disabled'
)
```

pfecht (approved):

PR #510, opened 2022-02-07
feat(elements|ino-datepicker): add new property `attachToBody`

BenPag:
## Proposed Changes

- add new property to set `flatpickr` `static` attribute
- change `flatpickr` target from `ino-input > div` to `ino-input`

## Things to check


packages/elements/src/components.d.ts
```diff
@@ -317,6 +317,10 @@ export namespace Components {
           * If true, an *optional* message is displayed if not required, otherwise a * marker is displayed if required
          */
         "showLabelHint"?: boolean;
+        /**
+          * Position the calendar inside the wrapper and inside the ino-datepicker.
+         */
+        "static"?: boolean;
```
Sl1nd, line 323: Probably it isn't necessary to provide an api to control the desired behavior. Thinking about the datepicker is suppose there will be no use case where a non static usage of the datepicker is useful. What do you think?

BenPag: I do not know if there is no use case for this. I will asking around if there is an use case for this.

Sl1nd: A possible issue is rendering the overlay above other content in deep nested scenarios. IMO it is necessary to control this behavior in favor of usage at different DOM-context

PR #518, opened 2022-02-14
fix(elements|ino-datepicker): reattach on DOM update

BenPag:
## Proposed Changes

- Support reattaching of datepicker by calling `this.create()` in `connectedCallback`

## Things to check

### DIFF
--- a/packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
+++ b/packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
@@ -299,6 +299,7 @@ export class Datepicker implements ComponentInterface {
       minDate: this.min,
       maxDate: this.max,
     });
+    this.create();
   }
 
   componentDidLoad() {


packages/elements/src/components/ino-datepicker/ino-datepicker.tsx
```diff
@@ -299,6 +299,7 @@ export class Datepicker implements ComponentInterface {
       minDate: this.min,
       maxDate: this.max,
     });
+    this.create();
```
Sl1nd, line 302: According to stencil component lifecycle https://stenciljs.com/docs/component-lifecycle this will process flatpickr creation twice on component initialization. Maybe the solution is on consumer level to prevent reattachement of the component?

janivo: That's right, the `create` method will be called twice on initialization but the first time will not result in the creation of the datepicker as `target` will be empty:

In line 318 or 319:
```tsx
const target = this.el.querySelector('ino-input > div') as HTMLElement;

if (this.disabled || !target) {
      return;
}
```

janivo (approved):

PR #529, opened 2022-02-21
fix(elements|ino-markdown-editor): handle change of `initialValue`

BenPag:
## Proposed Changes

- add new watcher function for property `initialValue`

## Things to check

### DIFF
--- a/packages/elements/src/components/ino-markdown-editor/ino-markdown-editor.tsx
+++ b/packages/elements/src/components/ino-markdown-editor/ino-markdown-editor.tsx
@@ -48,9 +48,7 @@ export class MarkdownEditor implements ComponentInterface {
 
   @Watch('initialValue')
   handleInitialValueChange(newInitialValue: string): void {
-    this.editor.commands.setContent(this.markdownToHtml(newInitialValue), true);
-    this.textareaRef.value = this.htmlToMarkdown();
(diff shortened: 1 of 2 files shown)


packages/elements/src/components/ino-markdown-editor/ino-markdown-editor.tsx
```diff
@@ -86,9 +93,7 @@ export class MarkdownEditor implements ComponentInterface {
   componentDidLoad(): void {
     this.createEditor();
     if (this.initialValue) {
-      this.editor.commands.setContent(this.markdownToHtml(), true);
-      this.textareaRef.value = this.htmlToMarkdown();
-      this.textareaRef.rows = this.textareaRef.value.split('\n').length;
+      this.handleInitialValueChange(this.initialValue);
```
janivo, line 96: IMO it's kinda strange that we call a handler here. We should extract the content of `handleInitialValueChange` into another function (maybe `initializeEditor`) and call it.

janivo (changes requested):

packages/elements/src/components/ino-markdown-editor/ino-markdown-editor.tsx
BenPag: Dr. Strange was working here^^ 🌀 I´ll change it,

janivo (approved):

PR #530, opened 2022-02-21
chore: add react-outputtarget

### LINKED ISSUE #404: Remove findDOMNode from React Wrapper
### Bug Report

When using react bindings with create-react-app and TypeScript there are some deprecation errors within the console.

### Is this a regression?

No.

Did this behavior use to work in the previous version?

- [X] I don't know
- [ ] Yes, the previous version in which this bug was not present was: `vX.Y.Z`
- [ ] No

### Minimal Reproduction

Render any components with the newest React version (eg. `inoList`).

### Expected behavior

There are no deprecation warnings.

### Screenshots

image

### Your Environment

| Software            | Version |
| ------------------- | ------- |
| @inovex.de/elements |    4.0.0     |
| Browser             |    any     |
| Operating System    |    any     |

### Possible solutions

Update internal React version. See https://github.com/inovex/elements/blob/e93b20f7ac05500a11dbbeefbbcd382da6146ef3/packages/elements-react/src/components/createComponent.tsx#L37-L40

JCofman:
Closes #404

## Proposed Changes
- adds official react-output-target implementation by stencil to generate react components

## Things to check

### DIFF
--- a/packages/elements-react/package.json
+++ b/packages/elements-react/package.json
@@ -48,8 +48,8 @@
     "@trust/webcrypto": "^0.9.2",
     "@types/jest": "^26.0.14",
     "@types/node": "^14.17.0",
-    "@types/react": "^16.14.21",
-    "@types/react-dom": "^16.9.14",
+    "@types/react": "^17.0.39",
+    "@types/react-dom": "^17.0.11",
     "@types/react-router": "^4.4.5",
     "@typescript-eslint/eslint-plugin": "^4.14.2",
     "@typescript-eslint/parser": "^4.14.2",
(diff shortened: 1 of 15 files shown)


packages/elements-react/package.json
```diff
@@ -59,18 +59,14 @@
...
     "shx": "^0.3.2",
-    "typescript": "^4.1.6"
+    "typescript": "^4.5.5"
   },
   "peerDependencies": {
-    "react": "^16.9.0",
-    "react-dom": "^16.9.0",
-    "react-router": "^4.3.1",
-    "react-router-dom": "^4.3.1"
+    "react": "^17.0.2",
```
janivo, line 68: This would break for projects using `react` < 17.0.0, wouldn't it?

packages/elements-react/src/components/index.ts
```diff
@@ -1,216 +1,61 @@
...
-import { defineCustomElements } from '@inovex.de/elements/dist/loader';
-
-import { createReactComponent } from './createComponent';
+/* eslint-disable */
+/* tslint:disable */
+/* auto-generated react proxies */
+import { createReactComponent } from './react-component-lib';
 
-defineCustomElements(window);
-addIcons(ICON_PATHS);
```
janivo, line 11: I suppose that this file is auto-generated. Is there a way to keep the `addIcons(...)` function?

janivo (commented):
Hey Jacob, thank you very much for your contribution! It's really appreciated.
I've only got two points we should consider.

I suppose that the files you added are taken from the stencil package, right?

janivo:
One more thing: if you rename the branch to start with `elements-` the pipeline will work

JCofman:
moved to new PR with elements prefix branch name

PR #531, opened 2022-02-22
chore: add react-outputtarget

JCofman:
Closes https://github.com/inovex/elements/issues/404

## Proposed Changes

- adds official react-output-target implementation by stencil to generate react components

## Things to check

### DIFF
--- a/packages/elements-react/package.json
+++ b/packages/elements-react/package.json
@@ -48,8 +48,8 @@
     "@trust/webcrypto": "^0.9.2",
     "@types/jest": "^26.0.14",
     "@types/node": "^14.17.0",
-    "@types/react": "^16.14.21",
-    "@types/react-dom": "^16.9.14",
+    "@types/react": "^17.0.39",
+    "@types/react-dom": "^17.0.11",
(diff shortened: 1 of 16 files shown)


janivo:
Almost, the name of the branch has to start with `elements-` 😅

packages/elements-react/package.json
```diff
@@ -59,18 +59,14 @@
     "eslint-plugin-prefer-arrow": "^1.2.3",
     "eslint-plugin-react": "^7.22.0",
     "np": "^3.1.0",
-    "react": "^16.9.0",
-    "react-dom": "^16.9.0",
-    "react-router": "^4.3.1",
-    "react-router-dom": "^4.3.1",
+    "react": "^17.0.2",
```
silentHoo, line 62: There's a breaking change in here we need to document: The React version changes from 16 -> 17 which breaks a few things for the consuming applications.

JCofman:
Ok, I am actually not sure what I do wrong I have created several new branches with specific names starting `elements-`. Is there a reason we have introduced the rule and where can I find it 😅

janivo:
see #537

PR #538, opened 2022-02-24
chore(elements): enable source-maps in dev mode

janivo:
This PR enables source maps that are generated by stencil. 
This feature has been part of the [2.9 release of stencil](https://ionicframework.com/blog/announcing-stencil-v2-9/).

I've made a quick comparison and the build size has more than doubled with this feature enabled. So I restricted the generation of source maps to the `dev` script. 

Here's a quick before/after comparison:

Before:
image

After:
image


packages/elements/package.json
```diff
@@ -43,7 +43,7 @@
   "scripts": {
     "build": "stencil build --docs --legacy && node scripts/generateWrappedIcons.js",
     "clean": "shx rm -rf dist",
-    "dev": "stencil build --docs --watch --legacy",
+    "dev": "NODE_ENV=development stencil build --docs --watch --legacy",
```
JCofman, line 46: Nice I didn't know about the sourceMap Feature. If we want to make sure that env settings work also on Windows environment we may use one of the following: 

* https://github.com/kentcdodds/cross-env
* https://github.com/toddbluhm/env-cmd

JCofman (dismissed):

packages/elements/package.json
janivo: You are right. I'll change it to use `cross-env` 👍🏼

JCofman (approved):
LGTM 👍

PR #539, opened 2022-02-25
chore: remove fonts hosted by google

### LINKED ISSUE #532: Replace landing page font source
### Refactor Request

We already implemented using static.inovex.de as font source. Now we want to use the host as well for the landing page.

If the font is missing at static.inovex.de we need to upload it first. Then replace the font urls within these files:

* https://github.com/inovex/elements/blob/master/packages/storybook/src/stories/docs/styleguide/styles.scss
* https://github.com/inovex/elements/blob/master/packages/landingpage/src/views/WelcomeIndex.vue

janivo:
Closes #532 

## Proposed Changes

- Remove the fonts hosted by Google in the following packages:
  - `landingpage`
  - `storybook`
- No replacement is needed because the fonts were not in use

### DIFF
--- a/packages/landingpage/src/views/WelcomeIndex.vue
+++ b/packages/landingpage/src/views/WelcomeIndex.vue
@@ -149,7 +149,6 @@ export default {
 
 <!-- Add "scoped" attribute to limit CSS to this component only -->
 <style scoped>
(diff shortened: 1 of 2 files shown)


BenPag:
Do we not need to replace the font source with `static.inovex.de`? Or is this already done?

BenPag (approved):

PR #547, opened 2022-03-03
chore(elements): add stencil linter

janivo:
Stencil now has its own linter, so I decided to add this one.

## Proposed Changes

- add rules
- remove unused/unnecessary rules
- fix warning/errors

### DIFF
--- a/packages/elements-angular/elements/src/directives/proxies.ts
+++ b/packages/elements-angular/elements/src/directives/proxies.ts
@@ -756,7 +756,7 @@ export class InoNavItem {
 
 export declare interface InoOption extends Components.InoOption {
   /**
-   *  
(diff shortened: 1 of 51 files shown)


BenPag:
Do you installed this package as peer dependency? By documentation of `stencil-eslint` this is remommended.

https://github.com/ionic-team/stencil-eslint#installation

janivo:
@BenPag no I did not. The docs state that the installation is only necessary if we use NPM pre version 7.0.0 which ships with Node V15.0.0.

Maybe we should rise the recommended node version? Our current recommendation (based on `package.json` `engine` and the `.nvmrc` file) is `v14.17`

pfecht (approved):

PR #555, opened 2022-03-16
fix(elements|ino-datepicker): fix datepicker hiding after input

BenPag:
## Proposed Changes

- apply `this.inline` check on leading slot

## Things to check


pfecht (approved):
Looks good, approved if you add this info to the comment in l.24 (see line 25)

janivo (approved):

PR #562, opened 2022-03-18
chore: add svgo to minimize icons and autoadd currentColor

JCofman:
## Proposed Changes

- adds SVGO to optimize the size of icons after commit and automatically sets `fill="currentColor`

size-reduction


janivo (changes requested):
Very nice Jacob, thank you 👍🏼 
One thing: I think I would move the `svgo.config.js` to the root of `elements/` as it's not really part of the package and rather a configuration.

packages/elements/src/assets/ino-icon/code_block.svg
```diff
@@ -1,14 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="22.521" height="20.409" viewBox="0 0 22.521 20.409">
```
BenPag, line 1: Compared by rich diff this icon is now broken.

BenPag (changes requested):

packages/elements/src/assets/ino-icon/code_block.svg
janivo: Can confirm, can be seen here: https://elements.inovex.de/unicorn/elements-icons-chore-auto-optimize-icons/?path=/docs/graphic-ino-icon--all-icons

BenPag (approved):

pfecht (dismissed):

janivo (approved):

Issue #574, opened 2022-04-19
Refactor stories to follow unified story format


silentHoo:
@janivo did some research and came up with the following PR to simplify our stories: https://github.com/inovex/elements/pull/490

All stories should follow the <ino-checkbox> example story.

This way of writing our stories allows us:
* to use typings in our stories which results in fewer errors
* to write less because we can reuse templates
* to include our description in the stories instead of having a README and a story for each component

PR #595, opened 2022-05-10
refactor(storybook|ino-radio): reformat stories

MariaLStefan:
Part of #574 

### Proposed Changes
- reformat stories of ino-radio to follow unified story format

### DIFF
--- a/packages/storybook/src/stories/ino-radio/ino-radio.stories.ts
+++ b/packages/storybook/src/stories/ino-radio/ino-radio.stories.ts
@@ -58,13 +58,3 @@ export const Playground = template.generatePlaygroundStory();
 export const Unchecked = template.generatePlaygroundStory()
 export const Checked = template.generateStoryForProp('checked', true);
 export const Disabled = template.generateStoryForProp('disabled', true);
-
-/**
- * Checks and disables this element.
- */
-export const CheckedAndDisabled = template.generatePlaygroundStory();
-
-CheckedAndDisabled.args = {
-  checked: true,
-  disabled: true,
-};


packages/storybook/src/stories/ino-radio/ino-radio.stories.ts
```diff
@@ -14,56 +15,57 @@ export default {
...
+  args => html`
+    <ino-radio
+      checked="${args.checked}"
+      disabled="${args.disabled}"
+      name="${args.name}"
+      value="${args.value}"
+    >
+      Radio Button Label
+    </ino-radio>
+  `
```
BenPag, line 51: In case of using the template string the intends generates an not well formated example code.

image

```suggestion
  args => html`
  <ino-radio
    checked="${args.checked}"
    disabled="${args.disabled}"
    name="${args.name}"
    value="${args.value}"
  >
    Radio Button Label
  </ino-radio>
`
```

packages/storybook/src/stories/ino-radio/ino-radio.stories.ts
```diff
@@ -2,6 +2,7 @@ import { Components } from '@inovex.de/elements';
 import { useEffect } from '@storybook/client-api';
 import { Meta, Story } from '@storybook/web-components';
```
BenPag, line 3: I think you can remove hiere the `Story` import statement.

BenPag (changes requested):
Nice work and really thanks for updating the stories 👍

packages/storybook/src/stories/ino-radio/ino-radio.stories.ts
pfecht: https://github.com/storybookjs/storybook/issues/12022#issuecomment-859540813 Can this help?

janivo: Yes, it's working as described but disabling the decorators also disables all of our event handlers as they are also implemented as decorators

Issue #453, opened 2021-11-04
Stories for Custom-Properties (CSS-Vars)


janivo:
It would be nice to be able to test all CSS-Vars we provided for each component in the corresponding storybook page.
We could use [custom matcher](https://storybook.js.org/docs/react/essentials/controls#custom-control-type-matchers) to use color-pickers, for example.

PR #596, opened 2022-05-11
fix(elements|ino-icon-button): improve button styling in disabled state

### LINKED ISSUE #513: ino-icon-button: buggy disabled state (filled)
### Bug Report

Filled  `ino-icon-button`  perpend primary background color in disabled state.

grafik

As a consumer I suppose the BG-color will be inverted to a more decent color (like grey). In general it would be nice to provide a custom property to customize this behavior.

**Acceptance Criteria**
* Add disabled color style (configurable via CSS Variable)
* Change default background color (should match design)

AlessaRad:
Closes #513 

## Proposed Changes

- specify styling for disabled state
- add CSS properties for button background and icon in disabled state
- add exemplary CSS properties story

### DIFF
--- a/packages/storybook/src/stories/ino-icon-button/ino-icon-button.stories.ts
+++ b/packages/storybook/src/stories/ino-icon-button/ino-icon-button.stories.ts
@@ -119,6 +119,25 @@ export const States = () => html`
           activated
         ></ino-icon-button>
       </div>
+
+      <div class="flex-child">
+        <h4>Disabled</h4>
+        <ino-icon-button
+          icon="time"
+          color-scheme="primary"
+          disabled
+        ></ino-icon-button>
+      </div>
+
+      <div class="flex-child">
+        <h4>Disabled (filled)</h4>
+        <ino-icon-button
+          icon="time"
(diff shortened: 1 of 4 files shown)


packages/storybook/src/stories/ino-icon-button/ino-icon-button.stories.ts
```diff
@@ -166,6 +185,73 @@ export const Variations = () => html`
...
+      disabled="${args.disabled}"
+      filled="${args.filled}"
+      color-scheme="primary"
+      activated="${args.activated}"
+      icon="time"
+    >
+    </ino-icon-button>
+  `;
+};
+CSSProperties.args = {
```
pfecht, line 213: @janivo this might be something we could also address in the template generator. Would be nice if we could: 
 * extract the name of the css vars from the config of the element
 * define a set of CssVarTypes (PxBased, Color) so that we only have to say which css has what type.

We don't need to address this here, but lets add some more background to https://github.com/inovex/elements/issues/453 then.

janivo (approved):

PR #598, opened 2022-05-12
refactor(storybook|ino-select): reformat stories

MariaLStefan:
Part of #574 

### Proposed Changes
- reformat stories of ino-select to follow unified story format


janivo (approved):
nice work!

PR #599, opened 2022-05-12
refactor(storybook|ino-switch): reformat stories

MariaLStefan:
Part of #574 

### Proposed Changes
- reformat stories of ino-switch to follow unified story format

### DIFF
--- a/packages/storybook/src/stories/ino-switch/ino-switch.stories.ts
+++ b/packages/storybook/src/stories/ino-switch/ino-switch.stories.ts
@@ -2,6 +2,7 @@ import { Components } from '@inovex.de/elements';
 import { useEffect } from '@storybook/client-api';
 import { Meta, Story } from '@storybook/web-components';
 import { html } from 'lit-html';
+import { TemplateGenerator } from '../template-generator';
 import { decorateStoryWithClass, withColorScheme } from '../utils';
 import './ino-switch.scss';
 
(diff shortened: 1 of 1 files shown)


packages/storybook/src/stories/ino-switch/ino-switch.stories.ts
```diff
@@ -38,27 +46,49 @@ export const Playground: Story<Components.InoSwitch> = (args) => html`
...
   <ino-switch disabled>Disabled</ino-switch>
-`;
+`);
+/**
+ * Change the state of the `<ino-switch>` element by setting the following properties
+ * 
+ * - `checked`: Marks this element as checked. (**unmanaged**)
+ * - `disabled`: Disables this element.
+ */
+export const States = templateStates.generatePlaygroundStory();
```
janivo, line 93: Instead of showcasing all states, IMO it might be enough to only show the `disabled` state like this:
```
export const Disabled = template.generateStoryForProp('disabled', true);
```

packages/storybook/src/stories/ino-switch/ino-switch.stories.ts
```diff
@@ -27,9 +28,16 @@ export default {
       return story();
     },
   ],
-} as Meta;
+  args: {
+    checked: false,
```
janivo, line 32: Think it would be nicer if the default state is switched on
```suggestion
    checked: true,
```

janivo (changes requested):

PR #600, opened 2022-05-17
feat(elements|ino-icon-button): use slotted icons

### LINKED ISSUE #512: ino-icon-button: support custom icons
### Feature Request

Currently it is not possible to provide custom icons to `ino-icon-button` component. Only existing icons could be used via the `icon` property.

One idea is to provide an additional `src` property. Similar to the `ino-icon` component custom svgs could be adressed from consumers assets folder.

Sl1nd:
Closes #512 

## Proposed Changes

- Move api to use slotted icons
- Adjust Storybook
- Add deprecation warnings  

## Things to check

### DIFF
--- a/packages/elements/src/components.d.ts
+++ b/packages/elements/src/components.d.ts
@@ -486,6 +486,7 @@ export namespace Components {
         "filled"?: boolean;
         /**
           * The name of the icon of this element.
+          * @deprecated This property is deprecated and will be removed with the next major release. Instead, use the `icon-leading` slot.
          */
         "icon"?: string;
         /**
@@ -2194,6 +2195,7 @@ declare namespace LocalJSX {
         "filled"?: boolean;
         /**
           * The name of the icon of this element.
+          * @deprecated This property is deprecated and will be removed with the next major release. Instead, use the `icon-leading` slot.
          */
         "icon"?: string;
         /**
--- a/packages/elements/src/components/ino-icon-button/ino-icon-button.tsx
+++ b/packages/elements/src/components/ino-icon-button/ino-icon-button.tsx
@@ -12,9 +12,13 @@ import {
   Watch,
 } from '@stencil/core';
 import classNames from 'classnames';
+import { hasSlotContent } from '../../util/component-utils';
 
 import { ColorScheme, ButtonType } from '../types';
 
+/**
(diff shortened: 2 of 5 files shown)


packages/elements/src/components.d.ts
```diff
@@ -2194,6 +2195,7 @@ declare namespace LocalJSX {
         "filled"?: boolean;
         /**
           * The name of the icon of this element.
+          * @deprecated This property is deprecated and will be removed with the next major release. Instead, use the `icon-leading` slot.
```
pfecht, line 2198: Do we really want to remove it? I'd say we keep both. The prop is very convenient for many use-cases and the slot provides flexibility for edge cases.

packages/elements/src/components/ino-icon-button/ino-icon-button.tsx
```diff
@@ -12,9 +12,13 @@ import {
   Watch,
 } from '@stencil/core';
 import classNames from 'classnames';
+import { hasSlotContent } from '../../util/component-utils';
 
 import { ColorScheme, ButtonType } from '../types';
 
+/**
+ * @slot icon-leading - For the icon to be prepended
```
pfecht, line 20: How about the default slot? Can we check if the default slot exists?

pfecht (changes requested):
Looks good in general, just two comments on the deprecation and the slot behaviour.

packages/elements/src/components/ino-icon-button/ino-icon-button.tsx
```diff
@@ -134,7 +140,8 @@ export class IconButton implements ComponentInterface {
           type={this.type}
         >
           <div class="mdc-icon-button__ripple"/>
-          <ino-icon icon={this.icon} class="mdc-icon-button__icon" />
+          {this.icon && !iconSlotHasContent && <ino-icon icon={this.icon} class="mdc-icon-button__icon" /> }
```
janivo, line 143: `this.icon ? <ino-icon ...> : <slot>`

janivo, line 143: Wrap with:
```
<span class="mdc-icon-button__icon">
  // ...
</span>
```

janivo (approved):

PR #632, opened 2022-06-07
refactor(storybook|ino-table-header-cell): reformat stories

MariaLStefan:
Part of #574 

### Proposed Changes
- reformat stories of ino-table-header-cell to follow unified story format


packages/storybook/src/stories/ino-table-header-cell/ino-table-header-cell.stories.ts
```diff
@@ -40,9 +42,19 @@ export default {
       return story();
     },
   ],
-};
+  args: {
+    autofocus: true,
+    notSortable: false,
+    label: 'Label, e.g. Belegart',
```
BenPag, line 48: I think we should use an english  example and not in german. ;)

BenPag (changes requested):

janivo (approved):

PR #633, opened 2022-06-07
refactor(storybook|ino-table): reformat stories

MariaLStefan:
Part of #574 

### Proposed Changes
- reformat stories of ino-table to follow unified story format


packages/storybook/src/stories/ino-table/ino-table.stories.ts
```diff
@@ -23,113 +24,123 @@ export default {
...
-export const Playground = args => html`
+const tableContent = html`
+    <tr slot="header-row"><ino-table-header-cell column-id="id" sort-start="asc"  label="ID"></ino-table-header-cell>
+      <ino-table-header-cell column-id="name" sort-start="asc" label="Name"></ino-table-header-cell>
+      <ino-table-header-cell column-id="release" label="Release Date"></ino-table-header-cell>
+      <ino-table-header-cell column-id="box-office" label="Box Office"></ino-table-header-cell>
+      <ino-table-header-cell column-id="rating" label="Rating" not-sortable></ino-table-header-cell>
+      <ino-table-header-cell column-id="another1" label="Another Column" not-sortable></ino-table-header-cell>
+      <ino-table-header-cell column-id="another2" label="Another Column" not-sortable></ino-table-header-cell>
+    </tr>
```
BenPag, line 53: Can you take a look on that issue?
image

packages/storybook/src/stories/ino-table/ino-table.stories.ts
```diff
@@ -23,113 +24,123 @@ export default {
...
-}
-withSortDirection(Playground, 'sortDirection', 'asc')
-
+`);
 
-export const SelectionWithCheckboxes = () => {
+export const Playground = template.generatePlaygroundStory();
+export const Loading = template.generateStoryForProp('loading', true);
+export const NoHover = template.generateStoryForProp('noHover', true);
+export const SortColumnId = template.generateStoryForProp('sortColumnId', 'id');
```
BenPag, line 137: I think we should add an extra text information that the sorting function is not part of the `ino-table` and have to provided by consumer.

packages/storybook/src/stories/ino-table/ino-table.stories.ts
```diff
@@ -23,113 +24,123 @@ export default {
...
+export const SortDirection = template.generateStoryForProp('sortDirection', 'desc');
```
BenPag, line 138: What is the difference between `SortColumnId` and `SortDirection`? 

In the storybook it look like the same. Should me merge this two in one story?

BenPag (changes requested):

packages/storybook/src/stories/ino-table/ino-table.stories.ts
MariaLStefan: `sortColumnId` identifies the column by which the table should be sorted by and `sortDirection` tells us the direction (ascending or descending) of the sort. 
They are not the same because `sortDirection` needs a `sortColumnId` to know by which column the rest of the table should be aligned by. The template for both are the same, the description is the only difference. Therefore it makes sense to merge the two. I will do that.

packages/storybook/src/stories/ino-table/ino-table.stories.ts
MariaLStefan: I put the disclaimer in the description of the merged story `SortColumnIdAndSortDirection`

packages/storybook/src/stories/ino-table/ino-table.stories.ts
```diff
@@ -126,16 +126,26 @@ const template = new TemplateGenerator<Components.InoTable>(
...
 `);
 
 export const Playground = template.generatePlaygroundStory();
 export const Loading = template.generateStoryForProp('loading', true);
 export const NoHover = template.generateStoryForProp('noHover', true);
-export const SortColumnId = template.generateStoryForProp('sortColumnId', 'id');
-export const SortDirection = template.generateStoryForProp('sortDirection', 'desc');
+
+/**
+ * For table sorting there are to needed props:
```
BenPag, line 138: I think a typo issue is here

packages/storybook/src/stories/ino-table/ino-table.stories.ts
```diff
@@ -42,15 +42,15 @@ export default {
   },
 };
 
-const tableContent = html`
-    <tr slot="header-row"><ino-table-header-cell column-id="id" sort-start="asc"  label="ID"></ino-table-header-cell>
+const tableContent = 
+    html`<tr slot="header-row"></tr>
```
BenPag, line 46: Is this right, that the `tr` does not inclues the `ino-table-header-cell` elements

packages/storybook/src/stories/ino-table/ino-table.stories.ts
```diff
@@ -42,15 +42,15 @@ export default {
...
+    <ino-table-header-cell column-id="id" sort-start="asc"  label="ID"></ino-table-header-cell>
       <ino-table-header-cell column-id="name" sort-start="asc" label="Name"></ino-table-header-cell>
```
BenPag, line 48: Indents are not equal here

BenPag (changes requested):
only small things left ;)

packages/storybook/src/stories/ino-table/ino-table.stories.ts
MariaLStefan: if it includes the `ino-table-header-cell` elements the big gap appears... i couldn't find another way to refactor the template so that the gap disappears.  
It doesn't change the appearance of the table so i left it like that.

Issue #643, opened 2022-06-09
ino-radio-group: add option to change the direction


janivo:
### Feature Request

The `<ino-radio-group>` is horizontally oriented by default and cannot be changed. It should be possible to align the `<ino-radio>` button vertically.

PR #652, opened 2022-06-22
feat(elements|ino-radio-group): add option to align radios vertically & horizontally

TobiasHeimGalindo:
resolves: https://github.com/inovex/elements/issues/643


packages/elements/src/components/ino-radio-group/ino-radio-group.tsx
```diff
@@ -162,10 +169,29 @@ export class RadioGroup implements ComponentInterface {
...
+        break;
+    }
+    return style;
+  }
+
   render() {
     return (
       <Host>
-        <div class="ino-radio-group-wrapper" tabIndex={0}>
+        <div class={this.setAlignment()} tabIndex={0}>
```
janivo, line 194: You can use `classNames` to implement conditional classes like this:
```tsx
const classes = classNames({
	'ino-radio-group-wrapper': true,
  'ino-radio-group-wrapper--vertical': this.alignment === 'vertical',
	'ino-radio-group-wrapper--horizontal': this.alignment === 'horizontal',
})

// ...

	<div class={classes} tabIndex={0}>
```

And I would always set the `ino-radio-group-wrapper` as the default class.

packages/elements/src/components/ino-radio-group/ino-radio-group.scss
```diff
@@ -3,6 +3,12 @@ ino-radio-group {
     width: fit-content;
   }
 
+  .ino-radio-group-wrapper-vertical {
+    width: fit-content;
+    display: flex;
+    flex-direction: column;
+  }
+
```
janivo, line 11: I think I would prefer this way because it's more explicit:

```suggestion
    width: fit-content;
    display: flex;
  
  	&--vertical {
  	    flex-direction: column;
  	}
  	
  	&--horizontal {
  	    flex-direction: row;
  	}
	}

```

packages/elements/src/components/ino-radio-group/ino-radio-group.tsx
```diff
@@ -27,6 +31,10 @@ export class RadioGroup implements ComponentInterface {
    * If there is an ino-radio child with the given value, the radio-button will be checked and the other radio-buttons unchecked.
    */
   @Prop() value?: string | number | null;
+  /**
+   * sets the alignment of the radios
```
janivo, line 35:
```suggestion
   * Sets the alignment of the radios to either vertical or horizontal.
```

Issue #653, opened 2022-06-22
Deploy example projects and link them in the landingpage


pfecht:
Since we now have all our example projects in a single repository, we can simply use our existing deployment mechanisms to deploy the example projects.

Acceptence Criteria:
- The behaviour should be kept pretty simple (no versioning), similar to the deployment of the landingpage. 
- The example projects are linked in the landinpage

PR #670, opened 2022-07-26
ci: deploy example projects and link them to landingpage

MariaLStefan:
Closes #653 

- 3 new git workflows. Angular and Vue deployment works, React deployment contains bugs, could be fixed with latest vite update.
- New landingpage section for example-projects


packages/landingpage/src/components/IvxHeader.vue
```diff
@@ -25,7 +25,7 @@
           {{ version.name }} {{ i === 0 ? "(latest)" : "" }}
         </ino-option>
       </ino-select>
-
+
```
BenPag, line 28: Can you revert these changes? The extra spaces are not needed I guess :)

packages/elements-react-example/src/shared/components/NavItemLink.tsx
```diff
@@ -1,5 +1,8 @@
 import { Link, useMatch } from 'react-router-dom';
 import { InoIcon, InoNavItem } from '../InovexElements';
+// const InoIcon = import.meta.glob('./shared/InovexElements');
+// const InoNavItem = import.meta.glob('./shared/InovexElements');
```
BenPag, line 4: Why you comment this out?

.husky/_/husky.sh
```diff
@@ -0,0 +1,30 @@
...
+  sh -e "$0" "$@"
+  exitCode="$?"
+
+  if [ $exitCode != 0 ]; then
+    echo "husky - $hook_name hook exited with code $exitCode (error)"
+    exit $exitCode
+  fi
+
+  exit 0
+fi
```
BenPag, line 30: Do you need this? I removed this file, because we do not use husky anymore.

BenPag (changes requested):
Thank you for the integration 👍 Small things left ;)

Following is not part of your issue:
Recently I worked on the packages deployment process. Personally I try to optimize the wohle CI stuff. In this case I think we should discuss if we want to integrate the example projects deployment into the release flow. And check if we need an manual trigger for example project deployment.

.github/workflows/test.yml
```diff
@@ -35,26 +35,6 @@ jobs:
...
-            packages/elements-react/dist/
-
-    - if: github.ref == 'refs/heads/master'
-      name: upload elements-angular artifact
-      uses: actions/upload-artifact@v2
-      with:
-        name: elements-angular
-        path:
-            packages/elements-angular/elements/dist/
+    #- if: github.ref == 'refs/heads/master'
```
BenPag, line 38: Do we want to upload build artifacts for non master branches? In which cases we need this?

packages/elements-react-example/vite.config.ts
```diff
@@ -6,5 +6,14 @@ export default defineConfig({
   plugins: [react()],
   server: {
     open: '/'
-  }
+  },
+  base: '/example-projects/react/',
```
BenPag, line 10: Is this config working locally too?

.github/workflows/test.yml
janivo: Whops, forgot to uncomment this one

packages/elements-react-example/vite.config.ts
janivo: Yes it does

BenPag (approved):

PR #694, opened 2022-08-04
ci: refactor workflow pipelines

### LINKED ISSUE #691: Refactor and CleanUp CI-Workflows
### Refactor Request

- clean up CI-Workfows
- use shared actions to remove duplicates 
- speed up piplines through caching and artifact download if possible

BenPag:
Closes #691

## Proposed Changes
- use shared actions
- improve depencency flow of angular packages


.github/workflows/artifacts-download/action.yml
```diff
@@ -2,29 +2,75 @@ name: 'Download elements build artifacts'
 description: 'Download elements build artifacts that can be used by release workflows'
 author: 'Benjamin Pagelsdorf'
 
+inputs:
+  workflow-id:
+    description: 'ID of workflow to download artifacts from'
+    required: false
+    default: 'true'
+  download-core-artifacts:
+    description: 'Do download core artifacts?'
```
janivo, line 11:
```suggestion
    description: 'Download core artifacts?'
```

.github/workflows/artifacts-download/action.yml
```diff
@@ -2,29 +2,75 @@ name: 'Download elements build artifacts'
...
+    required: false
+    default: 'true'
+  download-example-artifacts:
+    description: 'Do download example project artifacts?'
```
janivo, line 15:
```suggestion
    description: 'Download example project artifacts?'
```

.github/workflows/deploy-example-projects.yml
```diff
@@ -0,0 +1,28 @@
...
+          workflow-id: ${{ github.event.workflow_run.workflow_id }}
+          download-core-artifacts: false
+          download-example-artifacts: true
+
+      - name: Deploy Example Projects to Github Pages 🚀
+        uses: JamesIves/github-pages-deploy-action@4.1.7
+        with:
+          branch: gh-pages # The branch the action should deploy to.
+          commit-message: Deploying example projects
+          target-folder: example-projects
```
janivo, line 28: These changes needs to be merged with the changes in the current master

.github/workflows/artifacts-upload/action.yml
```diff
@@ -2,52 +2,74 @@ name: 'Upload elements build artifacts'
...
+    required: false
+    default: 'true'
+  upload-storybook-artifact:
+    description: 'Do upload storybook artifact?'
+    required: false
+    default: 'true'
+  upload-example-artifacts:
+    description: 'Do upload example project artifacts?'
+    required: false
+    default: 'true'
```
janivo, line 17: IMO the default values should be set to `false`. That way you have to explicitly set these values which make for better readability. 

e.g.:
This is better readable:
```
    - uses: ./.github/workflows/artifacts-upload
      with:
	upload-storybook-artifact: true
```
than this:
```
    - uses: ./.github/workflows/artifacts-upload
      with:
        upload-core-artifacts: false
        upload-example-artifacts: false
```

janivo (changes requested):
Overall really nice job. Have a look at my remarks and rebase/merge onto the current master.

.github/workflows/install/action.yml
```diff
@@ -1,27 +0,0 @@
-name: "Install"
-description: "Installs the inovex-elements packages"
```
janivo, line 2: In comparison to the deleted `install` script, the Node-Binaries do not seem to be cached in the `install-dependencies` script, which results in the pipeline being approx. 30 sec slower. Could you have a look at that?

Before:
image

After:
image

janivo (changes requested):

.github/workflows/install/action.yml
BenPag: Compared to the old `install` routine, we are now using the node version specified in `.nvmrc`.  I think GitHub caches only the latest node major versions which is `v14.20.0`. In my point of view we have following solution to fix this:

- not using the `.nvmrc` in pipline and allow any version of `v14.x`
- update version in `.nvmrc` to '14.20.0' (is for example 14.x allowed in this file?)

janivo: Let's try to use `v14` in our `nvmrc`. According to [this comment](https://github.com/nvm-sh/nvm/issues/995#issuecomment-506818398) it should work just fine

BenPag: It seems this works, but Windows systems this feature is not supported. On Windows `nvm` wants to use `v14.0.0`.

PR #695, opened 2022-08-09
style(elements|ino-checkbox): update design

### LINKED ISSUE #681: ino-checkbox: Apply new Design
Apply the new design of our designers to the `<ino-checkbox>` component without modifying its API.

MariaLStefan:
Closes #681 

Applies new styling to ino-checkbox.

Color variables not set jet.

### DIFF
--- a/packages/elements/src/components/ino-checkbox/ino-checkbox.scss
+++ b/packages/elements/src/components/ino-checkbox/ino-checkbox.scss
@@ -4,12 +4,12 @@
 @use '@material/checkbox';
 @use 'base/theme';
 
-$primary-dark: theme.color(primary, dark);
-$primary-base: theme.color(primary);
-$primary-light: theme.color(primary, light);
-$grey: theme.color(light);
-$grey-light: theme.color(light, light);
-$white: theme.color(primary, contrast);
+$primary-dark: #2C02FF;
+$primary-base: #4B4EFF;
+$primary-light: #2D5DFF4D;
+$grey: #ABB1CC;
+$grey-light: #CBD0E0;
+$white: #fff;
 
 $unchecked-color: $grey;
 $checked-color: $primary-dark;
--- a/packages/elements/src/components/ino-checkbox/ino-checkbox.scss
+++ b/packages/elements/src/components/ino-checkbox/ino-checkbox.scss
@@ -4,54 +4,26 @@
 @use '@material/checkbox';
 @use 'base/theme';
 
-:host {
-  /**
-   * @prop --ino-checkbox-color: primary color of the checkbox
-   * @prop --ino-checkbox-color-hover: color of the checkbox on hover
-   * @prop --ino-checkbox-color-active: color of the checkbox if active
-   * @prop --checkbox-container-color-unchecked: color of the checkbox container if unchecked
-   * @prop --ino-checkbox-text-color-disabled: color of the checkbox text if disabled
-   * @prop --ino-checkbox-background-color-disabled: color of the checkbox background if disabled
-   * @prop --ino-checkbox-checkmark-color: primary color of the check mark
-   * @prop --ino-checkbox-checkmark-color-disabled: color of the check mark if disabled
-   */
+$primary-dark: #2C02FF;
+$primary-base: #4B4EFF;
+$primary-light: #2D5DFF4D;
+$grey: #ABB1CC;
+$grey-light: #CBD0E0;
+$white: #fff;
 
-   --checkbox-color: var(--ino-checkbox-color, #{theme.color(primary)});
-   --checkbox-color-hover: var(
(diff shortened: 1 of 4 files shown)


packages/elements/src/components/ino-checkbox/ino-checkbox.scss
```diff
@@ -16,38 +16,39 @@
...
-  --checkbox-background-color-disabled: var(
-    --ino-checkbox-background-color-disabled,
-    #{theme.color(light, light)}
-  );
-  --checkbox-checkmark-color: var(--ino-checkbox-checkmark-color, #fff);
-  --checkbox-checkmark-color-disabled: var(
-    --ino-checkbox-checkmark-color-disabled,
-    #{theme.color(light)}
-  );
+   --checkbox-color: var(--ino-checkbox-color, #{theme.color(primary)});
```
janivo, line 19: I don't think that we need these variables anymore. We should use the SASS-Vars for our primary color but that's it.

janivo:
The indeterminate mark should be a little slimmer IMO. Maybe `65%` width?
image

packages/elements/src/components/ino-checkbox/ino-checkbox.scss
```diff
@@ -57,52 +58,38 @@
...
+  }
+
+  .mdc-checkbox__checkmark {
+    inset: 2.3px;
+    width: 15px;
   }
 
   .mdc-checkbox--disabled ~ label {
-    color: var(--checkbox-text-color-disabled);
+    color: #CBD0E0;
```
janivo, line 76: Can be moved to a variable since we use this value twice

packages/elements/src/components/ino-checkbox/ino-checkbox.scss
```diff
@@ -57,52 +58,38 @@
...
+  .mdc-checkbox__background {
+    --mdc-checkbox-unchecked-color:#4B4EFF;
+    
   }
 }
 
-:host(:active) .mdc-checkbox:not(.checked),
+//Styling during click, pressed state
+:host(:active) .mdc-checkbox:not(.checked), :host(:focus),
 :host(active) .mdc-checkbox:not(.mdc-checkbox--disabled) {
```
janivo, line 90: Typo
```suggestion
:host(:active) .mdc-checkbox:not(.mdc-checkbox--disabled) {
```

packages/elements/src/components/ino-checkbox/ino-checkbox.scss
```diff
@@ -16,38 +16,39 @@
...
 }
 
+
 :host {
-  --mdc-checkbox-unchecked-color: var(--checkbox-container-color-unchecked);
-  --mdc-checkbox-checked-color: var(--checkbox-color);
-  --mdc-checkbox-disabled-color: var(--checkbox-background-color-disabled);
+  --mdc-checkbox-unchecked-color: #ABB1CC;
+  --mdc-checkbox-checked-color: #2C02FF;
+  --mdc-checkbox-disabled-color: #CBD0E0;
```
janivo, line 51: Instead of using these variables, can't we use [these mixins](https://material.io/components/checkboxes/web#api)? I would prefer mixins because that's the "official" API

janivo (changes requested):
Great job overall, looks just like the design 👍🏼

packages/elements/src/components/ino-checkbox/ino-checkbox.scss
```diff
@@ -4,54 +4,26 @@
...
-   );
-   --checkbox-checkmark-color: var(--ino-checkbox-checkmark-color, #fff);
-   --checkbox-checkmark-color-disabled: var(
-     --ino-checkbox-checkmark-color-disabled,
-     #{theme.color(light)}
-   );
-}
+$primary-dark: #2C02FF;
+$primary-base: #4B4EFF;
+$primary-light: #2D5DFF4D;
```
janivo, line 9: Instead of strictly following the design, we might want to use our color set which is quite similar. You can find it here: https://github.com/inovex/elements/blob/master/packages/elements/src/components/base/_theme.scss

And use it like this:
`theme.color(primary)` or `theme.color(primary, light)`


This allows us to be flexible in the future, e.g. if we want to use completely different colors for all our components

janivo (changes requested):
Nicely done, just one last thing

packages/elements/src/components/ino-checkbox/ino-checkbox.scss
MariaLStefan: Thanks! I changed it to the theme colors, but primary light is currently a bit off. Should be fixed after we adjust the theme colors

janivo (approved):

PR #698, opened 2022-08-15
test(storybook): add snapshot tests

### LINKED ISSUE #713: Add snapshot tests
### Feature Request

To prevent regression e.g. when updating the MDC libraries, we should have a way to see if anything (e.g. the structure of a component) changed unintentionally.

Add snapshots tests.

### Description of the considered alternatives

An alternative is visual regression testing (a.k.a. screenshot tests). As #683 showed, these seem to be really flaky in our current setup and therefore are not suited for us at the moment.

### Additional context

Snapshot testing is already partially implemented by #698 but there is still a problem with components that generate a unique ID to function properly. These ids changes with every render and therefore differ every test run.

janivo:
Closes  #713 

- add snapshot test for all stories

⚠️  ~~currently blocked by https://github.com/storybookjs/test-runner/issues/68#issuecomment-1214989696~~


janivo:
Seems to be fixed for now. Probably some weird dependency errors.

But there is still one problem: In some of our components we generate random IDs which lead to following errors:
```
    -       <label for="ino-switch-id__4h67rw409">
    +       <label for="ino-switch-id__8g9ktqkwt">

```

We need to think about how we want to handle these cases. IDs are necessary for `<label>` elements.

PR #700, opened 2022-08-16
fix(elements|ino-datepicker): run initial validation before load

janivo:
Fixes this warning which occurs on every render:
image


MariaLStefan (approved):
Looks good, waring doesn't show up again.

PR #705, opened 2022-08-19
style(elements|ino-segment-button): update design

### LINKED ISSUE #690: ino-sgement-button(-group): Apply new Design
Apply the new design of our designers to the component without changing its API.

TobiasHeimGalindo:
Closes #690 

## Proposed Changes

- update the style of the ino-segment-button & ino-segment-group

**remove depricated css properties:** _--ino-segment-button-checked-shadow_ & _--ino-segment-button-hover-shadow_

**add 2 new css properties to accomodate new design:** _--ino-segment-button-focus-border_ & _--ino-segment-button-pressed-color_

## TODO:

- [x] wrap a border around the children of the **_ino-segment-group_**
- [x] delete comments & redundant css variables (_$left-segment-border_...)


packages/elements/src/components/ino-segment-button/ino-segment-button.scss
```diff
@@ -64,10 +84,7 @@ $left-right-segment-border: 0 24px 24px 24px;
       background-color: #fff;
     }
   }
-
-  .mdc-button--outlined {
-    @include button.outline-width(1px);
-    @include button.outline-color(var(--segment-button-color));
+  .mdc-button--unelevated {
     @include button.ink-color(var(--segment-button-color));
```
janivo, line 88: In comparison, the font style looks different from the design. 
image

- The letters should not be capitalized
- The text should have no letter spacing (at least not as wide)
- The background should be slightly brighter (#E8EDFF)

packages/elements/src/components/ino-segment-button/ino-segment-button.scss
```diff
@@ -76,21 +93,22 @@ $left-right-segment-border: 0 24px 24px 24px;
     }
 
     &:hover {
-      @include button.outline-color(var(--segment-button-hover-color));
-      @include button.ink-color(var(--segment-button-hover-color));
-      box-shadow: var(--segment-button-hover-shadow);
+      @include button.container-fill-color(var(--segment-button-hover-color));
+      @include button.ink-color(var(--segment-button-checked-color));
+    }
+    &:focus {
```
janivo, line 99: elements-ino-segment-button-tabbing
When tabbing through the group, the font stays greyish. According to the design (`focused` state).it should be blue.

packages/elements/src/components/ino-segment-button/ino-segment-button.scss
```diff
@@ -54,8 +58,24 @@ $left-right-segment-border: 0 24px 24px 24px;
...
+    );
+    &:hover {
+      border-radius: $round-border;
+      @include animation.border-radius(
+        ino-segment-button-edged-to-round,
+        $edged-border,
+        $round-border,
+        0.15s
+      );
+    }
```
janivo, line 77: You probably took that from the `ino-button` right?
I think it would be a nice idea, to extract these styles into a custom file. Maybe create a new scss file in `src/components/base/common.scss` and provide a mixin that is imported by the `ino-button` and `ino-segment-button`?

packages/elements/src/components/ino-segment-button/ino-segment-button.tsx
```diff
@@ -119,7 +119,7 @@ export class InoSegmentButton implements ComponentInterface {
 
     const buttonClasses = classNames({
```
janivo, line 120: We should disable the shadow root here as well.

janivo (changes requested):
Overall, nice work! Just some minor things.

Also, maybe it might be worth it, to drop the MDC-Button and just implement these states by ourselves as we don't really have much complexity. The usage of the mdc button rather creates complexity by using the respective classes/mixins.

janivo:
According to the design, the default segment button should be of `44px` height

janivo (changes requested):

packages/elements/src/components/ino-segment-button/ino-segment-button.scss
```diff
@@ -1,96 +1,87 @@
...
+      0.15s,
+      $round-to-edged: true
+    );
+
+    min-height: 44px;
+    line-break: auto;
+    line-height: 100%;
+
+    cursor: pointer;
     @include typography.font($size: s, $weight: none);
```
janivo, line 48: One thing left: the font should be medium according to the design. I think that should be `medium` instead of `none`.

janivo (changes requested):

PR #712, opened 2022-08-25
style(elements|ino-range): apply new design

MariaLStefan:
Closes #687

PR #717, opened 2022-08-30
refactor(elements|ino-switch): remove color-scheme prop

### LINKED ISSUE #707: ino-switch: Remove color-scheme property
No real use case exists to support such a property. It rather leads to confusion why there is such a property for the switch but not for other components.

TobiasHeimGalindo:
Closes #707

## Proposed Changes

-removes the color-scheme property of the ino-switch component


packages/elements/src/components/ino-switch/ino-switch.scss
```diff
@@ -7,35 +7,39 @@
...
+@mixin set-track-border-color {
   .mdc-switch__track {
     border: 1.5px solid white;
     margin: -1.5px;
-    box-shadow: 0 0 0 5px $color;
+    box-shadow: 0 0 0 5px $focus;
   }
 }
 
 @each $color in theme.$color-schemes {
```
MariaLStefan, line 22: We can remove this selector because we don't support all the theme colors anymore

MariaLStefan (changes requested):
Looks good! Just one thing left

packages/elements/src/components/ino-switch/ino-switch.scss
```diff
@@ -7,60 +7,62 @@
 $n-5: #abb1cc;
 $n-6: #cbd0e0;
 
-@mixin set-track-border-color($color) {
+/* colors */
+$primary: #2c02ff;
+$focus: #2d5dff4d;
```
janivo, line 12: I would suggest copying the `_new-theme.scss` from here https://github.com/inovex/elements/pull/712/files#diff-b6776d33e431be01d3f60db81bb96bd67c5766209cc39a44a003594c9b3fd943 and then using them instead of defining the colors again.

janivo (changes requested):

janivo:
One thing I noticed: The height of the disabled and non-disabled `ino-switch` is minimally different. See:
elements-ino-switch-disabled-jump

We should set a fixed height (the height of the disabled one), to prevent these jumps. Could you do that in this PR as well?

janivo (dismissed):

PR #748, opened 2022-09-21
ci: make several improvements of the release pipeline

### LINKED ISSUE #747: Refactor Release-Pipeline
While trying to release a new elements version via the new release pipeline, there are some things I noticed that we should fix:
- The Github-Pages Branch is not correctly updated (see https://github.com/inovex/elements/commit/18f6d55d66e39028e422ddf151772be11abee08c), the tag should be used as a new version name (e.g. `v.7.1.2`) instead of the branch name (`elements-release-v-7.1.2`)
- When running `yarn publish:ci` lerna bumps the version again instead of working with the already bumped version. We should replace the lerna command with simple `npm publish` commands in each core package (elements, elements-angular, ...)
- Github Release
  - The Github Release is currently created before the version has been released to NPM. It should be created afterward
  - The [Release Action](https://github.com/ncipollo/release-action) has an `artifacts` option. We should use it to link our artifacts
  - The Changelog of the Github Release should be identical to the content of the `CHANGELOG.md` file, there is a mismatch right now

janivo:
Closes #747 

Taken from the issues:
- [x] Update Github-Pages Branch correctly
- [x] Run `npm publish` for each package instead of using `publish:ci`
- [x] Storybook version should be updated after the release (currently that's not working for some reason)
- [ ] Github Release
  - [x] Is created after the packages have been published
  - [ ] Artifacts are linked to the Github Release
  - [x] ~~The Changelog of the Github Release is identical to the content of our Changelog file~~

### DIFF
--- a/.github/workflows/deploy-version.yml
+++ b/.github/workflows/deploy-version.yml
@@ -64,15 +64,9 @@ jobs:
         working-directory: packages/elements-vue
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-      - name: Create release artifacts
-        run: npm pack @inovex.de/elements
-             && npm pack @inovex.de/elements-angular
-             && npm pack @inovex.de/elements-react
-             && npm pack @inovex.de/elements-vue
       - name: Create a GitHub release
         uses: ncipollo/release-action@v1
         with:
-          artifacts: ./*.tgz
           tag: ${{ steps.tag_version.outputs.new_tag }}
           name: Release ${{ steps.tag_version.outputs.new_tag }}
           body: ${{ steps.tag_version.outputs.changelog }}
--- a/.github/workflows/deploy-version.yml
+++ b/.github/workflows/deploy-version.yml
@@ -48,25 +48,21 @@ jobs:
         run: npm publish
         working-directory: packages/elements
         env:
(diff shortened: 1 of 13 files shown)


.github/workflows/deploy-version.yml
```diff
@@ -34,30 +34,44 @@ jobs:
...
+        run: npm publish
+        working-directory: packages/elements
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: Publish @inovex.de/elements-angular to NPM
+        run: npm publish
+        working-directory: packages/elements-angular
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
BenPag, line 55: Do we need this token? I think the npm token should be enough.

BenPag:
The current used GitHub Action does not support excluding changelog types. We can try to open an PR to support this feature and/or fork this action to get earlier access to this new feature.

janivo:
> The current used GitHub Action does not support excluding changelog types. We can try to open an PR to support this feature and/or fork this action to get earlier access to this new feature.

Ok, I think we should just try to generate similar changelogs. I updated the configuration to show a heading for style changes

PR #754, opened 2022-09-29
refactor: wrap icons on manual trigger

janivo:
Due to problems with the icon wrapper, I have hard-coded the wrapper so that a dynamic wrapping of the icons becomes obsolete. Adding new icons now only requires running the `integrate:icons` script.


BenPag (changes requested):

janivo:
Good points @BenPag 👍🏼 
I applied your feedback

packages/elements/scripts/integrateIcons.ts
```diff
@@ -0,0 +1,102 @@
...
+      '}',
+    ].join('\n') + '\n';
+
+  fs.writeFileSync(DST_ESM, finalFile);
+}
+
+const DST_ESM_D_TS = path.join(SRC_SVG_DIR, 'index.esm.d.ts');
+
+function makeIndexEsmD(iconData: SvgMetadata[]) {
+  const iconTypes = iconData.map(({ name }) => `\t'${name}': string;`);
```
BenPag, line 73: In our prettier settings we are using a tab-width of 2. So should we better use here two blanks instead of `\t`?

packages/elements/scripts/integrateIcons.ts
```diff
@@ -0,0 +1,102 @@
...
+function makeIndexEsm(svgMetadata: SvgMetadata[]) {
+  const imports = svgMetadata.map(
+    ({ filename, importName }) =>
+      `import ${importName} from '${BASE_PATH}/${filename}';`
+  );
+
+  const iconPathsProps = svgMetadata.map(
+    ({ name, importName }) => `\t'${name}': ${importName},`
+  );
+  const exports = svgMetadata.map(({ importName }) => `\t${importName},`);
```
BenPag, line 54: In our prettier settings we are using a tab-width of 2. So should we better use here two blanks instead of `\t`?

BenPag (approved):

PR #767, opened 2022-10-18
ci: run unicorn deployment by workflow_call trigger

BenPag:
## Proposed Changes

- use `workflow_call` trigger 
- call unicorn deployment from `test.yml`
- clean up / simplify cache logic
- provide deployment link in pull request

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,3 +39,9 @@ jobs:
         upload-core-artifacts: true
         upload-example-artifacts: true
         upload-storybook-artifact: true
+
+  unicorn:
+    needs:
+      - test
+      - build
+    uses: ./.github/workflows/deploy-unicorn.yml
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -39,3 +39,9 @@ jobs:
         upload-core-artifacts: true
         upload-example-artifacts: true
         upload-storybook-artifact: true
+
+  unicorn:
+    needs:
+      - test
```
BenPag, line 45: Should we want to provide an unicorn deployment independenly from the `test` job?

janivo: Yea, it would be nice if we could also trigger it manually. Wouldn't it be just adding `manual` to the triggers?

BenPag: Should the `manual` trigger on `test,yml` or `deploy-unicorn,yml`? In case of  `deploy-unicorn,yml` we need to add a build step for the binaries or add an input for the specific `workflow_id`.

janivo: You are right, that seems to be kinda complicated. I think it's fine then, we don't need to able to run an independent job

janivo (approved):

PR #780, opened 2022-11-07
fix(elements-angular|ino-markdown-editor): prevent styles from being applied in wrong places

### LINKED ISSUE #761: elements-angular|ino-markdown-editor: selecting and applying styles does not work
### Bug Report

When trying to style an already written line of text within the markdown-editor, the styles seem to be applied to the very last line not matter which one is selected. This behaviour does not always occur.

### Screenshots

angular-markdown-editor

### Additional context

Seems to occur only within the context of Angular ⚠️

BenPag:
Closes #761 

## Proposed Changes

- removing extra ControlValueAccessor for `ino-markdown-editor`
- add usage example of `ino-markdown-editor` in angular example project


BenPag:
ToDo: 
- [ ] Check if initial value triggers value change in angular reactive form

janivo (approved):

PR #784, opened 2022-11-09
feat(landingpage): add components section

### LINKED ISSUE #758: Add "the components" section on landingpage
Add "The components"  section on the landingpage

Image

TobiasHeimGalindo:
Closes #758


packages/landingpage-new/pages/home/component-sample/component-sample.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    setChip(false);
+
+                    setTimeout(() => {
+                      setChip(true);
+                    }, 5000);
+                  }}
+                >
+                  Removable Chip
+                </InoChip>
+              )}
```
janivo, line 58: IMO the removable variant is not suited for showing off the chips functionality because it can be irritating if the chip suddenly disappears. Maybe we should [use the filter functionality](https://elements.inovex.de/version/v7.1.2/?path=/docs/buttons-ino-chip--playground#filter) instead? Two chips with a short label (e.g. Pizza, Burger) should be enough

image

packages/landingpage-new/pages/home/component-sample/component-sample.tsx
```diff
@@ -0,0 +1,201 @@
...
+                  action-text="Some Action"
+                  timeout={5000}
+                  type="info"
+                  stay-visible-on-hover="false"
+                  onHideEl={() => {
+                    setSnackbar(false);
+                  }}
+                >
+                  This is a message
+                </InoSnackbar>
```
janivo, line 110: IMO the snackbar is also not really suitable because it makes element appear outside of our boxes/cards. Maybe we should just use the `<ino-button>` instead?

packages/landingpage-new/pages/home/component-sample/component-sample.tsx
```diff
@@ -0,0 +1,201 @@
...
+              )}
+            </div>
+            <div slot="footer" className={styles.footer}>
+              <div className={styles.component}>Snackbar</div>
+              <div className={styles.label}>Notification</div>
+            </div>
+          </InoCard>
+        </div>
+        <div className={styles.input}>
+          <InoCard disableElevation={true}>
```
janivo, line 120: I would suggest extracting the card in its own component, so we dont have to write `disableElevation={true}` on every card.

packages/landingpage-new/pages/home/component-sample/component-sample.module.scss
```diff
@@ -0,0 +1,84 @@
+@use '../../../../elements/src/components/base/typography';
```
janivo, line 1: You can not use these styles because they are defined in an separate packages and are currently not exported. Until we've made them available, we have to defined these styles ourselves. I would suggest creating a `typography.scss` in `styles/` and write some mixins. E.g.:

```scss
mixin light-font() {
		font-weight: ...;
		font-size: ...;
} 
```

This way, we can easily swap these out later.

janivo (changes requested):
Overall great job! One thing we should take into account is the responsiveness of this layout.
With a smaller screen, the cards should shrink to a defined `min-width` and then wrap to a `2x4` layout. On really small screens (e.g. mobile) the layout should transform into `1x8` matrix.
image

PR #821, opened 2022-12-14
refactor(landingpage): style and route corrections home

### LINKED ISSUE #820: Style and route corrections of the home page
Image

and update routing to new version (docs --> library; learn --> getting started)

ninaschlz:
Closes #820 

## Proposed Changes

- header: library instead of docs; getting-started instead of learn
- routing: refactored subroutes as we discussed in miro
- hyperlink "inovex elements" routes to /library
- rename "show documentation" button to "getting started" and route to /getting-started
- framework icons below ino-chip and success message to the right
- "contact us" button refering to /home#contact


packages/landingpage-new/components/home/header/install-chips.module.scss
```diff
@@ -1,15 +1,18 @@
 .container {
   display: flex;
-  column-gap: 20px;
+  row-gap: 20px;
+  flex-direction: column;
   position: relative;
 }
 
 .chipContainer {
-  width: 410px;
```
janivo, line 8: image

I think the chip has to be slightly wider. Also, the copy message should be moved to the right so it doesn't overlap with the chip. Could you also animate the confirmation message? For example, we could make it slide from behind the button. Should be easy to do it like this:
```scss
.successMessage {
  position: absolute;
  color: #68ce6a;
  font-weight: bolder;
  top: -5px;
  left: 0;
  transition: left 1s;

  &Show {
    left: 400px;
  }
}
```

Also necessary:
- increase z-index of chip container
- give chip-container a white background color
- give chip-container witdh `fit-content`

The only thing that's tricky is that the `left` position of the message needs to account for the width of the React chip.

janivo (changes requested):

packages/landingpage-new/components/home/header/install-chips.module.scss
ninaschlz: i removed the absolute position of the success message and placed it next to the chip. is the animation to the right important, or is it ok with visibility?

janivo (approved):

packages/landingpage-new/components/home/header/install-chips.module.scss
janivo: its fine like this 👍🏼

PR #855, opened 2023-01-11
fix(landingpage): correct order and links of frameworks

### LINKED ISSUE #850: Getting Started: Autoselect correct tabs on submenu link click
- when clicking on a subitem in the footer, the corresponding framework should be selected in the tab bar
- framework order should be exactly the same as in the footer

ninaschlz:
Closes #850 

## Proposed Changes

- same order of frameworks in home, getting-started and footer
- fix links of getting-started subroutes in footer


packages/landingpage-new/components/layout/footer.tsx
```diff
@@ -5,17 +5,21 @@ import LinkItem from './linkItem';
...
-              url={`${url}#${subRouteUrl}`}
-              name={subRouteName}
-              isDense={true}
-            />
-          ))}
+      {Routes.map(({ name: mainRouteName, url: mainRouteUrl, subRoutes }) => (
+        <div key={mainRouteUrl} className={styles.col}>
+          <LinkItem url={mainRouteUrl} name={mainRouteName} />
+          {subRoutes.map(({ name: subRouteName, url: subRouteUrl }) => {
+            const subrouteDivider =
```
janivo, line 12: We might have other sub routes divided by `/` in the future so it would make sense to extract this logic in a function like this:

```tsx
const getDividerByMainRoute = (mainRoute: MainRoutes): '/' | '#' => {
  switch (mainRoute) {
    case MainRoutes.GETTING_STARTED:
      return '/';
    default:
      return '#';
  }
};
```

packages/landingpage-new/pages/getting-started/index.tsx
```diff
@@ -1,19 +1,19 @@
...
 /**
- * Forwards to react guide as default
+ * Forwards to javascript guide as default
  */
 const GettingStarted = () => {
   const { pathname, replace } = useRouter();
 
   useMount(() => {
-    replace(pathname + '/' + Framework.REACT);
+    replace(pathname + '/' + Framework.JS);
```
janivo, line 13: The react guide will be probably be the most popular one. That's why I think we should keep routing to the react one by default (but no need to change the order tho).

janivo (changes requested):

janivo (approved):

PR #858, opened 2023-01-12
feat(landingpage): add coreaspects and make it responsive

### LINKED ISSUE #848: Implement our Core-Aspects
Now that #789 is done, we need should implement the core aspect section of our homepage. The icons can be found in our shared drive. Keep mobile in mind.

ninaschlz:
Closes #848 

## Proposed Changes

- add core aspects images and text
- make it responsive


packages/landingpage-new/components/home/core-aspects/core-aspects.tsx
```diff
@@ -1,29 +1,47 @@
-import { InoIcon } from '@elements';
 import styles from './core-aspects.module.scss';
+import versatile from '@assets/coreaspect-versatile.svg';
+import consistent from '@assets/coreaspect-consistent.svg';
+import futureproof from '@assets/coreaspect-futureproof.svg';
+import Image from 'next/image';
 
 export default function CoreAspects() {
   return (
     <div className={styles.secondContainer}>
```
janivo, line 9: I would choose another name instead of `secondContainer`. What about `coreAspectsContainer`?

packages/landingpage-new/components/home/core-aspects/core-aspects.tsx
```diff
@@ -1,29 +1,47 @@
...
       <div>
```
janivo, line 10: I also would give these its own class instead of relying on the element selector. How about `coreAspects`?

packages/landingpage-new/components/home/core-aspects/core-aspects.module.scss
```diff
@@ -5,13 +5,60 @@
...
   border-radius: 267px 0px;
+  column-gap: 40px;
 
   div {
-    padding-top: 122px;
+    height: 350px;
     display: flex;
     flex-direction: column;
     align-items: center;
+    justify-content: start;
```
janivo, line 15: The `start` value has a mixed browser support. We should use `flex-start` instead

janivo (changes requested):

janivo (approved):

PR #870, opened 2023-01-26
chore(landingpage): deploy landingpage on changes in landingpage dir

### LINKED ISSUE #813: Deploy new Landingpage on every push
- We are currently manually deploying our new landingpage project via our https://github.com/inovex/elements/actions/workflows/deploy-landingpage-new.yml action
- Landingpage should rebuild and deploy on every master push IF anything changed in `packages/landingpage-new`
- Keep the manual trigger

Resources: 
- https://www.techielass.com/trigger-a-github-action-workflow-when-a-folder-changes/
- https://stackoverflow.com/a/70711156
- https://github.com/marketplace/actions/changed-files

TobiasHeimGalindo:
Closes #813 

Currently checks if there were changes made in the landingpage-new directory to deploy. Very similar to the Manual "Deploy Landingpage New" Workflow.

- Optional: might be able to be written differently with [GitHub Action - Changed Files](https://github.com/marketplace/actions/changed-files) `steps.changed-files.outputs.changed` to run each build only if there are changes. Example:

```
name: Build elements-react
        run: yarn build
        working-directory: packages/elements-react
        if: steps.changed-files.outputs.changed == 'true'
```


.github/workflows/deploy-landingpage-new-on-change.yml
```diff
@@ -0,0 +1,52 @@
+name: Deploy Landingpage New On Change
+
+on:
+  push:
+    branches: 
+      - master
+    paths:
+      - 'packages/landingpage-new/**'
```
janivo, line 8: The actions are still the same as in `deploy-new-landingpage` right? If that's the case, it might be sufficient to move the trigger (the stuff I highlighted) to the `deploy-new-landingpage` action. I think you can define many triggers for one action

janivo (changes requested):

PR #876, opened 2023-02-03
feat(elements): add ino-accordion component

### LINKED ISSUE #872: Implement the ino-accordion
Magda has designed an accordion component which we'll need for our FAQ in the new landingpage.

- The accordion should be implemented according (:badumts:) to the design (default, hover, ...)
- The accordion should have a property `expanded`
  - is of type `boolean`
  - defaults to `false`
  - controls weather the default or expanded state is shown
 - The accordion should have a property `title`
   - is of type `string`
   - defaults to `null`
   - a warning should be printed (e.g. in `componentDidLoad`) if this property is missing
 - The accordion should have a default slot where the expanded content can be provided
 -  The accordion should have a event (e.g. `expandedChange`)
    - that emits when the expand button has been clicked
    - that emits when the title has been clicked
    - that emits the new expanded state as its detail

TobiasHeimGalindo:
Closes #872


packages/elements/src/components/ino-accordion/ino-accordion.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return (
+      <div class={inoAccordionClasses}>
+        <div class="ino-accordion--header" onClick={() => this.toggleExpand()}>
+          {!this.expanded ? <ino-icon icon="add" /> : null}
+          {this.expanded ? <ino-icon icon="close" /> : null}
+
+          <span class="ino-accordion--title">{this.accordionTitle}</span>
+        </div>
+        {this.expanded ? (
+          <div class="ino-accordion--content">
```
janivo, line 61: The expand/collapse mechanism should be animated (e.g. by animating the `max-height` of the content to `0` when collapsed and to some unrealistic value `9999px` when expanded)

see: https://stackoverflow.com/a/8331169

packages/elements/src/components/ino-accordion/ino-accordion.tsx
```diff
@@ -0,0 +1,68 @@
...
+    this.expandedChange.emit(this.expanded);
+  }
+  render() {
+    const inoAccordionClasses = classNames({
+      'ino-accordion': true,
+      'ino-accordion--expanded': this.expanded,
+    });
+    return (
+      <div class={inoAccordionClasses}>
+        <div class="ino-accordion--header" onClick={() => this.toggleExpand()}>
```
janivo, line 54: The `cursor` should be set to `pointer` to indicate a clickable area

packages/elements/src/components/ino-accordion/ino-accordion.tsx
```diff
@@ -0,0 +1,68 @@
...
+          {!this.expanded ? <ino-icon icon="add" /> : null}
+          {this.expanded ? <ino-icon icon="close" /> : null}
```
janivo, line 56: I would suggest the characters `+` and `-` as a the toggle. Also it would be great if we could animate it like this: https://stackoverflow.com/a/38136283

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -0,0 +1,38 @@
...
+    @include typography.typo(body-m);
+
+    color: #4C4B52;
+  }
+
+  &--title {
+    padding-left: 20px;
+    @include typography.typo(title-m);
+
+    color: #4655ff;
```
janivo, line 28: The color of the title varies depending on its `expand` state:
image

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -0,0 +1,38 @@
+@use 'base/typography-new' as typography;
+@use 'base/new-theme' as theme;
+
+.ino-accordion {
+  &:hover {
+    background: #f6fafe;
```
janivo, line 6: We should use the variables of the colors instead if its hex code. If it doesn't exist in the `new-theme` we should add it (in this case `p-01`) and use it in our component

image

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -0,0 +1,38 @@
...
+    align-items: center;
+    padding-left: 16px;
+    min-height: 48px;
+  }
+
+  &--content {
+    padding: 0 52px 32px 52px;
+    @include typography.typo(body-m);
+
+    color: #4C4B52;
```
janivo, line 21: use variables (see above)

packages/elements/src/components/ino-accordion/ino-accordion.scss
janivo, line 28: use variables (see above)

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -0,0 +1,38 @@
...
+
+  &--title {
+    padding-left: 20px;
+    @include typography.typo(title-m);
+
+    color: #4655ff;
+  }
+
+  &--expanded {
+    background: #f6fafe;
```
janivo, line 32: use variables (see above)

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -0,0 +1,38 @@
...
+
+    &:hover {
+      background: #f6fafe;
```
janivo, line 35: use variables (see above)

janivo (changes requested):

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
+    color:  theme.$n-10;
   }
 
   &--title {
-    padding-left: 20px;
+    padding-left: 10px;
     @include typography.typo(title-m);
 
-    color: #4655ff;
+    color: #4C4B52;
```
janivo, line 29: Could you extract this as a variable at the top of the file?

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
+
+  .ino-accordion--content-wrapper {
+    max-height: 0;
+    overflow: hidden;
+    transition: max-height 0.2s ease-out;
+  }
+  
+  .ino-accordion--content-wrapper.expanded {
+    max-height: 1000px;
+    transition: max-height 0.4s ease-in;
```
janivo, line 52: It should suffice to declare the animation once on the class:

```suggestion
    transition: max-height 0.2s ease-in-out;
  }
  
  .ino-accordion--content-wrapper.expanded {
    max-height: 1000px;
```

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
 
+
 .ino-accordion {
   &:hover {
-    background: #f6fafe;
+    background: theme.$p-01;
   }
   border-radius: 24px;
 
   &--header {
```
janivo, line 11: Should be `__header` because it's another block inside the component. `--` should be used to indicate a state of a block (e.g. `ino-accordion__title--selected`).

We are not strictly following BEM but you can read more about it here: https://css-tricks.com/bem-101/

```suggestion
  &__header {
```

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
+    cursor: pointer;
     display: flex;
     align-items: center;
-    padding-left: 16px;
+    padding-left: 10px;
     min-height: 48px;
   }
 
   &--content {
```
janivo, line 19: should be `__` (see above)

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
-    padding: 0 52px 32px 52px;
+    padding: 0 56px 32px 52px;
     @include typography.typo(body-m);
-
-    color: #4C4B52;
+    color:  theme.$n-10;
   }
 
   &--title {
```
janivo, line 25: should be `__` (see above)

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
-    background: #f6fafe;
+    background: theme.$p-01;
 
     &:hover {
-      background: #f6fafe;
+      background: theme.$p-01;
     }
   }
+
+  .ino-accordion--content-wrapper {
```
janivo, line 44: should be `__` (see above)

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
+    max-height: 0;
+    overflow: hidden;
+    transition: max-height 0.2s ease-out;
+  }
+  
+  .ino-accordion--content-wrapper.expanded {
```
janivo, line 50: should be `ino-accordion__content-wrapper` and `ino-accordion__content-wrapper--expanded`

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
   }
 
   &--expanded {
-    background: #f6fafe;
+    background: theme.$p-01;
 
     &:hover {
-      background: #f6fafe;
+      background: theme.$p-01;
     }
```
janivo, line 41: should not be necessary because the values are the same

packages/elements/src/components/ino-accordion/ino-accordion.tsx
```diff
@@ -41,27 +41,41 @@ export class Accordion implements ComponentInterface {
...
+    this.expandedChange.emit(!this.expanded);
   }
   render() {
     const inoAccordionClasses = classNames({
       'ino-accordion': true,
       'ino-accordion--expanded': this.expanded,
     });
+
+    const inoAccordionContent = classNames({
+      visibility: this.expanded ? 'visible' : 'hidden',
```
janivo, line 53: should be handled by a class (e.g. `ino-accordion__content--visible`)

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
+    transition: max-height 0.2s ease-out;
+  }
+  
+  .ino-accordion--content-wrapper.expanded {
+    max-height: 1000px;
+    transition: max-height 0.4s ease-in;
+  }
+}
+
+.ino-accordion--header button {
```
janivo, line 56: We should use classes instead of tag selectors.

e.g.:

```suggestion
.ino-accordion__toggle-button {
```

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -1,38 +1,97 @@
...
+  left: 25%;
+  right: 25%;
+  height: 10%;
+  top: 45%;
+}
+
+
+.ino-accordion--header button.clicked span {
+  transform: rotate(90deg);
+}
```
janivo, line 92: use classes.

e.g.:
`ino-accordion__toggle-button--expanded` and `ino-accordion__toggle-button--collapsed`

janivo (changes requested):
Really nice job with the animations 🥳 
Only thing that's left is the css part 👇🏼

packages/elements/src/components/ino-accordion/ino-accordion.tsx
```diff
@@ -50,26 +50,34 @@ export class Accordion implements ComponentInterface {
...
+            }
+          >
+            <span class="toggle-icon"></span>
+            <span class="toggle-icon"></span>
           </button>
           <span
-            class={`ino-accordion--title ${this.expanded ? 'expanded' : ''}`}
+            class={`ino-accordion__title ${
+              this.expanded ? 'ino-accordion__title--expanded' : ''
+            }`}
```
janivo, line 73: I would suggest to define the conditional class like this: https://github.com/inovex/elements/blob/e414aefaef9218dc11acc59f863d6ae6233d6bea/packages/elements/src/components/ino-chip/ino-chip.tsx#L140-L164

packages/elements/src/components/ino-accordion/ino-accordion.tsx
```diff
@@ -50,26 +50,34 @@ export class Accordion implements ComponentInterface {
...
           >
             {this.accordionTitle}
           </span>
         </div>
         <div
-          class={`ino-accordion--content-wrapper ${
-            this.expanded ? 'expanded' : ''
+          class={`ino-accordion__content-wrapper ${
+            this.expanded ? 'ino-accordion__content-wrapper--expanded' : ''
```
janivo, line 80: see above

janivo (changes requested):

janivo (approved):

PR #912, opened 2023-03-23
feat: add new elements logo

### LINKED ISSUE #906: Integrate the new elements logo
Our logo got a new fresh look! ✨ 

Check where our current logo is used and replace it with the new one.

Request standard sizes (16x16, 32x32, ...) from design team

janivo:
Closes #906 

- add the new logo as a shared resource at the root of our repository
- replace every occurrence
- ensure that all projects always access the newest logo version


packages/storybook/config/manager-head.html
```diff
@@ -1,7 +1,12 @@
 <!DOCTYPE html>
 <html lang="en">
-  <link rel="shortcut icon" href="favicon.ico" />
-  <link rel="icon" type="image/png" href="favicon.png" sizes="192x192" />
+  <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
+  <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png">
+  <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png">
+  <link rel="manifest" href="/site.webmanifest">
```
BenPag, line 6: For unicorn deployments the icons are not shown, becaue of the `/` in the `href` attribute.

packages/elements-react-example/src/App.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { BrowserRouter as Router, Navigate, Route, Routes } from 'react-router-dom';
 import './index.scss';
-import logo from './favicon.svg';
+import logo from '../../../assets/logo/elements.svg';
```
BenPag, line 4: Should we define a `paths`  value for this to access these files like `@assets/logo/elements.svg` or `@logo/elements.svg`

janivo: I tried to extend the root `tsconfig` with the path mappings but vite does not seem to find the files then 🤔

BenPag: Okay, this could relate to the missing extension of the base `tsconfig` in the packages. But we can this optimize later too :)

PR #933, opened 2023-04-19
feat(elements|ino-chip): add clickable option

### LINKED ISSUE #851: ino-chip: add not-clickable mode
The `ino-chip` can be used as a button alternative but also as a label for some simple data. Currently, the chip is always clickable which makes it appear to be interactive which is not always the case. We should provide a property (e.g. `clickable`) that disabled all interactions (hover effects, cursor, ...).

MariaLStefan:
Closes #851

### DIFF
--- a/packages/elements/src/components/ino-chip/ino-chip.scss
+++ b/packages/elements/src/components/ino-chip/ino-chip.scss
@@ -17,8 +17,12 @@ $chip-color-font: theme.$primary;
 $icon-size: 16px;
 
 ino-chip {
-
   height: 24px; // default height
+  pointer-events: none;
+  
+  &.ino-chip--clickable {
+    pointer-events: all;
+  }
 
   // base styling
   .ino-chip-container {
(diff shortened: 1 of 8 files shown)


BingeCode (dismissed):

packages/elements/src/components/ino-chip/ino-chip.scss
```diff
@@ -17,8 +17,12 @@ $chip-color-font: theme.$primary;
 $icon-size: 16px;
 
 ino-chip {
-
   height: 24px; // default height
+  pointer-events: none;
+  
+  &.ino-chip--clickable {
+    pointer-events: all;
+  }
```
janivo, line 25: Since `clickable` defaults to `true` I would suggest introducing a class `ino-chip--not-clickable` which disabled `pointer-events` if the class is set. Makes up for less code and improves readability IMO

janivo (changes requested):

janivo (approved):

PR #934, opened 2023-04-20
feat(landingpage): improve react getting started guide

### LINKED ISSUE #864: Improve our Getting Started section
Magda went over our Getting Started section and designed some improvements, we should implement now. You can find the mockups in our figma library. Implement these.

- add Menu on the right
  - the menu shows the current section the user is viewing (keyword: `IntersectionObserver`)
  - the menu items are clickable and scroll to the respective menu
- add sections via hashes (e.g. `getting-started/react#properties`)
- adjust styling according to the design in figma
  - for the prop/event tables use screenshots (don't show everything, see gdrive)
  - show the finished app as a screenshot on the top of the guide

Inspiration: https://atlassian.design/foundations/color-new

BingeCode:
Closes #864

Implemented new Design according to Figma (including new floating menu)

TODO: Standardize margins, font-sizes, apply standards across all guides

EDIT: The german counterpart will be adapted to the new design once the current implementation has been approved.


package-lock.json
janivo, line 1: This can be deleted

packages/landingpage-new/components/shared/navigationMenu.tsx
```diff
@@ -0,0 +1,84 @@
+import { useEffect, useState } from 'react';
+import styles from './navigationMenu.module.scss';
+
+export function NavigationMenu() {
+  const [activeSection, setActiveSection] = useState('preparation');
```
janivo, line 5: Instead of using plain strings I would suggest introducing an enum to handle all the sections.

e.g.:
```ts
enum Sections {
  PREPARATIONS = 'preparations',
  PROPERTIES = 'properties',
  // ...
}
```

packages/landingpage-new/components/shared/navigationMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  useEffect(() => {
+    const observer = new IntersectionObserver(
+      (entries) => {
+        entries.forEach((entry) => {
+          if (entry.isIntersecting) {
+            setActiveSection(entry.target.id);
+          }
```
janivo, line 13: If we introduce an enum, we could check here whether the `target.id` is part of the given enum. If not, we should print a warning into the console as a hint for fellow developers, that they might have forgotten to at the section to their enum

packages/landingpage-new/components/shared/navigationMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+        });
+      },
+      { rootMargin: '-30% 0px -70% 0px' } // top, right, bottom, left margins around the root element's bounding box
+    );
+
+    const sections = document.querySelectorAll('article section[id]');
+    sections.forEach((section) => {
+      observer.observe(section);
+    });
```
janivo, line 22: Can be simplified:
```suggestion
    sections.forEach(observer.observe);
```

packages/landingpage-new/components/shared/navigationMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+          >
+            <a href="#finishing-touches" onClick={handleAnchorClick}>
+              Finishing Touches
+            </a>
+          </li>
+          <li className={activeSection === 'wrapping-up' ? styles.active : ''}>
+            <a href="#wrapping-up" onClick={handleAnchorClick}>
+              Wrapping up
+            </a>
+          </li>
```
janivo, line 79: This can be simplified to a single loop if we use an enum. Also, you could pass the respective section to the `handleAnchorClick` function instead of querying for it.

packages/landingpage-new/styles/globals.css
```diff
@@ -21,6 +22,10 @@ body {
   box-sizing: border-box;
 }
 
+p {
+  line-height: 25px;
+}
```
janivo, line 27: Why is this necessary?

packages/landingpage-new/styles/globals.css
```diff
@@ -32,3 +37,35 @@ ino-chip * {
...
+  grid-template-rows: auto 1fr;
+  grid-template-columns: minmax(400px, 960px) 1fr;
+}
+
+.hint {
+  color: #908da2;
+  font-size: 12px;
+  margin-top: 8px;
+  margin-bottom: 24px;
+}
```
janivo, line 71: These styles should not be part of the globals as they are kinda specific to the guides section

yarn.lock
janivo, line 1: This file should not change unless we did some changes to the dependencies

packages/landingpage-new/components/shared/navigationMenu.module.scss
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.navigationMenu .active {
+  color: theme.$p-05;
+}
+
+.navigationMenu > ul {
+  list-style-type: none;
+  padding: 0;
+}
```
janivo, line 41: I would generally prefer a class for the `ul` element instead of using a selector like this

packages/landingpage-new/components/shared/navigationMenu.module.scss
```diff
@@ -0,0 +1,51 @@
...
+
+.aside > .navigationMenu > h5 {
+  margin-bottom: 40px;
+  margin-top: 0;
+}
+
+.navigationMenu > ul > li {
+  position: relative;
+  margin-top: 32px;
+}
```
janivo, line 51: see above

janivo (changes requested):
Great job overall! Just some minor things

packages/landingpage-new/styles/globals.css
BingeCode: Because it's not the default in the elements typography, even though it's specified in the design docs and used per default. Needs to be updated in the typography (body-large)

packages/landingpage-new/styles/guide.css
janivo, line 1: These styles are located in the globals styles directory but are actually exclusively used by the guides. I would suggest creating small components (e.g. `<Hint>`, `<TableShadow>`, ...) where the styles are scoped to the component and can then be used in our mdx guides by importing them like a regular react component. This way we prevent the styles from leaking into other styles and also ensure that we share these styles (or components) within our guides

packages/landingpage-new/components/shared/navigationMenu.tsx
```diff
@@ -49,34 +58,24 @@ export function NavigationMenu() {
   }
 
   return (
-    <aside className={styles.aside}>
+    <aside className={styles.aside + ' menu'}>
       <nav className={styles.navigationMenu}>
         <h5>REACT GUIDE</h5>
```
janivo, line 63: This should be provided by props as it could also be used by the angular guide in the near future

janivo (changes requested):

packages/landingpage-new/styles/globals.css
```diff
@@ -1,11 +1,18 @@
...
 html,
 body {
   padding: 0;
   margin: 0;
   font-family: Lato, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen,
-  Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
+    Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, sans-serif;
+}
+
+:root h1.header-d2,
```
janivo, line 12: What is this for? 🤔

janivo (changes requested):

packages/landingpage-new/styles/globals.css
BingeCode: I think it's part of an earlier commit and I just auto-formatted the file with prettier

janivo (approved):

PR #944, opened 2023-05-08
style(landingpage): improve wrap of core aspects

janivo:
## Proposed Changes

- improve wrap behaviour of core aspects to reduce taken up space

### DIFF
--- a/packages/landingpage-new/components/home/core-aspects/core-aspects.module.scss
+++ b/packages/landingpage-new/components/home/core-aspects/core-aspects.module.scss
@@ -13,23 +13,18 @@
     flex-direction: column;
     align-items: center;
     justify-content: flex-start;
-    width: 300px;
     text-align: center;
     margin: auto 0;
-  }
-}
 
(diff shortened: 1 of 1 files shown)


packages/landingpage-new/components/home/core-aspects/core-aspects.module.scss
```diff
@@ -13,23 +13,18 @@
...
     margin: auto 0;
-  }
-}
 
-@media only screen and (max-width: 1600px) {
-  .coreAspectsContainer {
-    .coreAspect {
-      width: 250px;
+    @media only screen and (max-width: 1600px) {
+      max-width: 200px;
```
MariaLStefan, line 20: Could you change this `200px` in something like `14.286rem` or `15rem`?

In the [PR 945](https://github.com/inovex/elements/pull/945) I changed all `px` to `rem` and I think we should stay consistent with this unit throughout the landing page

MariaLStefan (changes requested):
Looks good! Just one little change is needed, see below.

PR #945, opened 2023-05-08
refactor(elements): make theme available for consumers

### LINKED ISSUE #925: Make our theme available for our consumers
Similar to what we did with typography (provide a stylesheet that can be imported and used), we should do the same with our theme. We have a defined color theme in our core project, which contains all our colors that can be useful for others (e.g. also our landing page). 

AC:
- Create a stylesheet that can be imported and used by users
- Make the colors available by CSS Variables
- Write a short guide (see [stylesheet story in storybook](https://elements.inovex.de/version/v8.0.0/?path=/docs/docs-stylesheet--page)), which describes how to import and use the said stylesheet.
- Use these colors in our new landingpage

MariaLStefan:
Closes #925 

- changes colors on landingpage-new to global variables
- changes px to rem on landingpage
- improves responsiveness on some landingpgage-new section


janivo (changes requested):
While I agree that `rem` is the overall superior unit, I don't believe that it should be used everywhere as a replacement for `px`. `rem` is a unit that is suited to make content more accessible (e.g. apply a larger font-size). But it doesn't always make sense to use it for dimensions (e.g. `width`) or spacing (e.g. `margins`) because the dimension/spacing might not always be related to the accessibility aspect.  For example, does it make sense to enlarge a `box-shadow` if the user preference has set a larger font size? I don't think so. That's why I think we should reconsider using `rem` everywhere and focus on the theming story in this PR for now.

The good thing is that you seem to have made these `px` to `rem` changes in Commit 347f4f6c08458c39274fe1559e06d7ff5e2a88a7 so we can easily revert it for now and test these changes in another branch.

PR #946, opened 2023-05-09
refactor: improve README.md (#919)

### LINKED ISSUE #919: Improve README.md
Our README contains a lot of text which might be a bit overwhelming when looking at the repository for the first time.

Use [this tool](https://readme.so/de) to freshen up our README.

Recommended sections:

Image

BingeCode:
Closes #919 

Showcase and Demo need to be added in later PR.


scripts/getContributors.js
janivo, line 1: Instead of using Puppeteer, we can use the API of Github directly. Have a look at this function: https://github.com/inovex/elements/blob/4e4be6b8a7167afe3272beb7c37493fb0cb77424/packages/landingpage-new/pages/%5Blang%5D/about/index.tsx#L25

janivo (changes requested):

scripts/getContributors.js
BingeCode: This is just a quick helper function that I wrote to extract the contributors from the repo. I used it in a separate repo to extract the image urls and profile names. I just saved it here so that it can be used manually in the future in case the contributor section needs to be updated. Although this can be done manually in the future since it isn't a lot of work to add a few contributors manually. We don't _really_ need it in the repo but I thought it might come in handy when you want to update a bulk of new contributors. Let me know how you want this to be managed.

janivo (dismissed):

janivo (approved):

PR #968, opened 2023-06-05
style(landingpage): fix stylistic issues

### LINKED ISSUE #967: Fix pre-release landing page issues

janivo:
Closes #967


packages/landingpage/pages/[lang]/library/components/index.tsx
```diff
@@ -10,28 +10,48 @@ import { LangContext } from '../../../../types/langContext';
...
+  // prevent scrolling of body while in storybook
+  useEffect(() => {
+    document.body.style.overflow = 'clip';
+    setHideFooter(true);
+
+    return () => {
+      document.body.style.overflow = 'initial';
+      setHideFooter(false);
+    };
+  }, []);
```
Comment (BingeCode) [inline packages/landingpage/pages/[lang]/library/components/index.tsx:34]: Why is this necessary? I couldn't tell a difference between the scroll behavior on master vs this branch

packages/landingpage/utils/context/UiContext.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export interface UiContextType {
+  hideFooter: boolean;
+  setHideFooter: (value: boolean) => void;
+}
+
+export const UiContext = createContext<UiContextType | null>(null);
+
+const UiContextProvider = ({ children }: { children: ReactNode }) => {
+  const [hideFooter, setHideFooter] = useState<boolean>(false);
```
BingeCode, line 11: Super minor nitpicky thing: For booleans, it is often helpful to use a prefix that denotes a binary condition. For example, `is`, `has`, `can`, `should` etc. f.e. `isFooterHidden`

packages/landingpage/components/layout/footer/footer.tsx
```diff
@@ -27,10 +27,6 @@ export default function Footer() {
             ))}
           </div>
         ))}
-        <div className={`${styles.col} ${styles.empty}`}></div>
-        <div className={`${styles.col} ${styles.empty}`}></div>
-        <div className={`${styles.col} ${styles.empty}`}></div>
-        <div className={`${styles.col} ${styles.empty}`}></div>
```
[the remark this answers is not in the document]
janivo: Great tip, than you 👍🏼

packages/landingpage/pages/[lang]/library/components/index.tsx
Comment (janivo) [reply packages/landingpage/pages/[lang]/library/components/index.tsx]: The problem can be reproduced in the most current deployment of our landingpage:
elements-scroll-landingpage

The problem is that the whole page will be scrollable as soon as you reach the max scroll distance of the page inside the iframe. That makes it kinda awkward to use as you have to be careful not to "over scroll" the storybook page.

PR #970, opened 2023-06-07
refactor: add dark version of the elements logo

### LINKED ISSUE #947: elements Logo im Dark-Mode

janivo:
Closes #947


BingeCode:
The viewbox in the svg file for the non-dark version is not set as cleanly as with the new dark version. Maybe we could fix that as well to make sure the sizing is the same across both versions. Should be easily fixable with a cleaner export from the original design file.

Issue #957, opened 2023-05-25
Implement the new ino-tooltip design


janivo:
- Implement the tooltip according to the design in Figma
- A property `colorScheme` exists with the possible values being one of `primary`, `light`, `dark`
- A property `showArrow` exists that displays/hides an arrow pointing to the attached element (see tippy.js)
- A header text can be provided (probably via Props) that is styled according to the design
- Stories are written for
  - Default
  - Color Schemes
  - Arrow
  - Header

janivo:
⚠ Start with #959

PR #974, opened 2023-06-12
refactor(elements|ino-popover): implement new design

### LINKED ISSUE #959: Implement the new ino-popover design
Unlike the popover, our tooltip is only intended to display text. However, since they share the same design of underlying colors, shapes and fonts, it is necessary to match the design of the popover to that of the tooltip.

- The colorScheme matches the design of the ino-tooltip
- Shapes, shadows, etc. match the design of the ino-tooltip while allowing the user to insert any content via slot

It might be helpful to create a shared stylesheet to prevent duplicate code.

MariaLStefan:
Closes #959


packages/elements/src/components/ino-popover/ino-popover.tsx
```diff
@@ -242,8 +249,10 @@ export class Popover implements ComponentInterface {
 
     const options: Partial<Props> = {
       allowHTML: true,
+      theme: this.colorScheme,
```
Sl1nd, line 252: Nice! IMO it's the better approach to use the documented way to adjust styling of tippy instances. Currently the ino-tooltip uses an custom approach to add stylings. It would be nice to be consistent  in codebase on handling Tippy instances. It would be great to refactor ino-tooltip  to inject stylings via theming as well.

packages/elements/src/components/ino-popover/ino-popover.tsx
```diff
@@ -111,9 +118,9 @@ export class Popover implements ComponentInterface {
 
   /**
    * Sets the color scheme of the popup
-   * Valid options include: 'primary', 'transparent'
+   * Valid options include: `light`, `dark` and `primary`.
    */
-  @Prop() colorScheme: 'primary' | 'transparent' =
+  @Prop() colorScheme: 'light' | 'dark' | 'primary' =
```
Sl1nd, line 123: The developer experience would be improved if we define consistent apis for ino-tooltip & ino-popover as well. Currently ino-tooltip provides 'light','primary','secondary' & 'transparent' color schemes. It would be nice if both components make use of the same set of options. This would also simplify maintaining a consistend design & implementation.

packages/elements/src/components/ino-tooltip/ino-tooltip.tsx
```diff
@@ -138,6 +138,7 @@ export class Tooltip implements ComponentInterface {
       delay: this.delay,
       placement: this.placement,
       trigger: this.trigger,
+      arrow: true,
```
Sl1nd, line 141: As i don't know the proper design requirements i'm not sure if we should expose this option to developers or if we should hardcode it. In generel it will also improve developer experience for consumers if we are consistent in API design of both ino-tooltip & ino-popover

packages/elements/src/components/ino-popover/ino-popover.tsx
MariaLStefan: Do you mean this approach [TippyJS: creating-a-theme](https://atomiks.github.io/tippyjs/v6/themes/#creating-a-theme)? Because I used this approach to create theme-stylings which resemble our designs in [_tooltip-base.scss line 41 to 56](packages/elements/src/components/base/_tooltip-base.scss). 

`this.colorScheme` returns the name of this themes (`light`, `dark` or `primary`) and applies it to the tippy Instance. 

`ino-tooltip` uses the same theme styling from this file, see [PR #975](https://github.com/inovex/elements/pull/975)

packages/elements/src/components/ino-tooltip/ino-tooltip.tsx
MariaLStefan: This change is already implemented In PR #975

packages/elements/src/components/ino-popover/ino-popover.tsx
MariaLStefan: I added a `TippyThemes` type 👍
Did the same change in PR #957

janivo (dismissed):

janivo (approved):

PR #975, opened 2023-06-12
refactor(elements|ino-tooltip): implement new design

### LINKED ISSUE #957: Implement the new ino-tooltip design

MariaLStefan:
Closes #957


Sl1nd:
@MariaLStefan Looks very good! The only improvement would be to align the preselected colorScheme values of storybook playgrounds of tooltip and popover. For InoPopover currently "light" is preselected. InoTooltip's playground sample uses "primary" colorScheme. At this consumers could assume a different color schmes of tooltips and popovers on first look.

Sl1nd (approved):

packages/elements/src/components/ino-tooltip/ino-tooltip.tsx
```diff
@@ -182,14 +194,18 @@ export class Tooltip implements ComponentInterface {
...
-      <Host class={hostClasses}>
+      <Host>
         <div class="ino-tooltip__composer" role="tooltip">
-          <div class="ino-tooltip__inner">{this.label ? this.label : <slot />}</div>
+          <div class="ino-tooltip__inner">
+            {this.label ? 
+              this.label 
+              : 
+              <div>
+                {this.headerText && <header>{this.headerText}</header>}
```
janivo, line 205: IMO the `header` should be independent of the existence of `label`.

Rougly like this:
```jsx
{header && <header> ... </header>}
{this.label ?? <slot />}
```

janivo (changes requested):
There are some conflicts that need to be resolved

janivo (approved):

PR #991, opened 2023-06-26
refactor(storybook): use new stylesheet for storybook

### LINKED ISSUE #965: Use new Stylesheet in our storybook
Our storybook should use our new stylesheet (especially the colors) in order to always stay up to date

in order to use our stylesheet in the storybook, have a look at [packages/storybook/config/preview-head.html](https://github.com/inovex/elements/blob/master/packages/storybook/config/preview-head.html)

MariaLStefan:
Closes #965

## Proposed Changes

- add & use new color theme in storybook

### DIFF
--- a/packages/storybook/config/manager-head.html
+++ b/packages/storybook/config/manager-head.html
@@ -12,23 +12,27 @@
 <!-- Custom styles for the main area -->
 <script>
   document.title = 'Showcase - inovex elements';
+  
+  document.body.classList.add("inovex-elements-typo");
+  document.body.classList.add("inovex-elements-theme");
+
 </script>
 <style type="text/css">
   div[role='menuitem'] {
-    color: #3d84f5;
+    color: var(--inovex-elements-p-4);
   }
 
   div[role='menuitem']:hover,
   a[class^='css-glamorous-routedlink']:hover {
-    color: #232527;
+    color: var(--inovex-elements-n-7);
   }
 
   a[class^='css-glamorous-routedlink'] {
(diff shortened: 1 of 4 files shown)


packages/storybook/config/manager-head.html
```diff
@@ -48,11 +50,12 @@
     display: none;
   }
 
+  
   .os-content .sidebar-subheading {
     /* Title Small */
     letter-spacing: 0.02px;
     font-weight: 600;
-    color: #CBD0E0; /* n-6 */
+    color: var(--inovex-elements-n-3);
```
BingeCode, line 58: Why did you pick n-3 over the previous n-6? It's quite a light contrast to the background. I think n-6 would be better, perhaps even higher - for accessibility and general readability.

n-3:
image

n-6:
image

packages/storybook/config/manager-head.html
```diff
@@ -70,56 +73,71 @@
     transition: 0.15s ease-in-out;
   }
   
+  .os-content .sidebar-item button[data-selected="true"] {
+    background-color: rgb(var(--inovex-elements-n-3)) !important;
```
BingeCode, line 77: Overriding styles with !important is always risky. It makes it difficult to override this style at a later point. Is there really no way to reach higher rule specifity with css selectors / import ordering?

BingeCode (changes requested):
Looks good overall, just needs a minor readability fix and perhaps another change.

janivo (approved):

PR #1020, opened 2023-07-14
refactor(elements|ino-autocomplete): update design

### LINKED ISSUE #1014: ino-autocomplete: Update design
When working on #984, Serap found that the `border-radius` of the popover of the ino-autocomplete is inconsistent with our current design. That should be adjusted to match the design found in Figma in order to be consistent within our own design system.

TobiasHeimGalindo:
Closes #1014

### DIFF
--- a/packages/elements/src/components/ino-autocomplete/ino-autocomplete.scss
+++ b/packages/elements/src/components/ino-autocomplete/ino-autocomplete.scss
@@ -9,7 +9,7 @@
   margin: 0.5rem 0 0 0;
   padding: 0;
   background-color: white;
-  box-shadow: 0px 0px 22px 0px rgba(93, 91, 105, 0.35);
+  box-shadow: 0px 0px 22px 0px rgba(new-theme.$n-10, 0.35);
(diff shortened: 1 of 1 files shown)


packages/elements/src/components/ino-autocomplete/ino-autocomplete.scss
```diff
@@ -9,8 +9,8 @@
   margin: 0.5rem 0 0 0;
   padding: 0;
   background-color: white;
-  box-shadow: 0 1px 5px #DBDBDB;
-  border-radius: 4px;
+  box-shadow: 0px 0px 22px 0px rgba(93, 91, 105, 0.35);
```
MariaLStefan, line 12: We should use our theme as much as possible in our components. 
Can you try out if one of our neutral colors (with the same `alpha 0.35`) look good?

MariaLStefan (changes requested):
The new design looks great!

Just one little change left.

MariaLStefan (approved):
nice 👌

PR #1024, opened 2023-07-19
feat(elements|ino-markdown-editor): edit links after inserting them

### LINKED ISSUE #981: ino-markdown-editor: Link-Button is not working
## Description
As a user, I want to add a link to a certain text in the `ino-markdown-editor`. 

What I expect:
1. Write a text
2. Select the text where I want to provide a link to
3. Click the link button (🔗)
4. I'm able to somehow provide a link to the selected text passage (usually handled by a popover)

## Current state
We currently got a link button that's missing the functionality above. Here's an example:
Image

## Acceptance criteria:
- The user is able to provide a link by clicking the link icon in the ino-markdown-editor

## More
Draw inspiration from Github, Gitlab, Jira, Gdocs, etc. (anything that has a rich text editor).

BingeCode:
Closes #981 and #1060

## Proposed Changes

- Added a editing dialog for inserting, editing and deleting links.

## Optional tweaks
- Autofocus text input when dialog opens
Autofocus is tricky, because it requires a rework of how we implement the dialog. The current implementation (with Material Dialog) tries to control focus. When programmatically changing focus, the dialog retakes control. I tried setting focus on the input using the autofocus attribute on ino-dialog, setting it programmatically at the end of the event loop (`Promise.resolve().then(() => element.focus());`) and setting it with timeout (`setTimeout(() => element.focus(), 100); `). It didn't work.

- Auto submit on enter
Auto submit likewise is tricky because it requires working with shadow DOM, attaching event listeners to the input in the shadow dom and managing event configuration. For such a small use case it is not worth the trouble to get this to work.

https://github.com/inovex/elements/assets/48513535/433188f0-f815-4f53-8e30-75e245cd34d9


packages/elements/src/components/ino-markdown-editor/ino-markdown-editor.tsx
janivo, line 1: Instead of opening the dialog on top of the whole app, I would suggest just overlaying the markdown-editor because this action only affects this part of the application. The `ino-dialog` may not be suitable then but given the suggestions I made above, it's probably for the best to just implement some custom overlay with an input field.

janivo (changes requested):
Really nice job overall but I'm not sure if the dialog is the right solution in the end. Have a look at my remarks. What do you think?

janivo (approved):

janivo (dismissed):

packages/elements/src/components/ino-icon-button/readme.md
janivo, line 1: This happens sometimes while working in dev mode. You should be able to restore it by running `yarn build`

janivo (changes requested):

janivo (approved):

janivo (approved):

PR #1033, opened 2023-07-31
chore(elements-angular): fix elements-angular release

### LINKED ISSUE #1009: Fix Angular release
The current Angular release is broken, we need to fix it asap.

BenPag:
Closes #1009 

## Proposed Changes

- introduce custom build helper script to provide ng-package top-level package.json
- all builds are now located at packages/<name>/dist


BenPag:
> I've tried to include the `@inovex.de/elements-angular` Package in another project but I could not get it to work properly. Did you manage to test it in an external project?

So if you tried it through the yalc publish process than its right, because this was broken. I fixed that and test it with the angular tour of heroes project and now it works for me under following conditions:

1. run `npm install` with ` --legacy-peer-deps` parameter
2. add `"skipLibCheck": true,` to `compilerOptions` of tsconfig.json

janivo (dismissed):

BenPag:
> > > I've tried to include the `@inovex.de/elements-angular` Package in another project but I could not get it to work properly. Did you manage to test it in an external project?
> > 
> > 
> > So if you tried it through the yalc publish process than its right, because this was broken. I fixed that and test it with the angular tour of heroes project and now it works for me under following conditions:
> > 
> > 1. run `npm install` with ` --legacy-peer-deps` parameter
> > 2. add `"skipLibCheck": true,` to `compilerOptions` of tsconfig.json
> > 
> > P.S.: The changed yalc process is not optimized yet, because nx rebuilds the core package in dependent project build. We can improve this later in a seperate story I guess
> 
> No, I've tried it with the default `npm link` command and ran into typing errors. That's probably why `skipLibCheck` is working for you because the ts-compiler doesn't care about library typings then.
> 
> Now that I think about it, it might be related to my test project being on angular 14 and our lib expecting angular 15. Let's merge this and see if the Canary version works.
> 
> Could you check this @BenPag? Preferably it should be working without the `skipLibCheck` flag.

The test project I used was on Angular 16RC1 🤔 an there I used the `skipLibCheck` flag and `--legacy-peer-deps`

janivo (approved):

PR #1034, opened 2023-08-02
refactor(elements|ino-accordion): optimize transition

TobiasHeimGalindo:
This PR refactors our ino-accordion component, drawing from a [blog post](https://keithjgrant.com/posts/2023/04/transitioning-to-height-auto/) to implement smoother transitions. By transitioning to height auto with grid, I have optimized the toggling of the expanded state. This results in cleaner code and improved user experience in my opinion as now we avoid the instant expanding of the accordion and fade in text solution from before.

Before & After:

https://github.com/inovex/elements/assets/81302108/edaaff25-05cc-4576-92a3-d10d481dc111


janivo (approved):
Smooth as butter 👌🏼

PR #1035, opened 2023-08-03
feat(elements|ino-nav-drawer): implement mobile nav-drawer

### LINKED ISSUE #1010: Implement a mobile version of our nav-drawer
In https://github.com/inovex/elements/issues/956 we created and updated the design, now we want to implement the nav-drawer.

* Update the design of the current nav-drawer
* Implement a mobile version of the nav-drawer according to the design
* Implementing different levels/hierarchies of nav items will be implemented in #1031 and won't be needed for this story

TobiasHeimGalindo:
Closes #1010 

https://github.com/inovex/elements/assets/81302108/c5c668e1-fad2-4a05-84a8-15f972206054


packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
```diff
@@ -67,9 +66,21 @@ export class NavDrawer implements ComponentInterface {
...
+   */
+  @Prop() isMobile?: boolean = false;
+
+  // set the variant based on the isMobile prop
+  get programmaticVariant(): NavDrawerVariant {
+    if (this.isMobile) {
+      return 'modal';
+    }
+    return this.variant;
+  }
```
MariaLStefan, line 83: You can use stencils `@Watch()` annotation to detect changes in `isMobile`. This function can then set the variant depending on the `isMobile` content.

```
  @Watch('isMobile')
  handleIsMobileChange() {
    if (this.isMobile) {
      this.variant = 'modal';
    }
  }
````
Further down the line, you only have to check `this.variant` to get according css classes

File: packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.scss [hunk 2]
```diff
@@ -223,4 +230,40 @@
...
+
+    & + .mdc-drawer-scrim {
+      backdrop-filter: blur(4.5px);
+      background: rgba(167, 183, 212, 0.41);
+    }
+  }
+
+  .visually-hidden {
+    position: absolute;
+    clip: rect(1px, 1px, 1px, 1px);
```
MariaLStefan, line 262: clip is deprecated https://developer.mozilla.org/en-US/docs/Web/CSS/clip
We should use the newer [clip-path](https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path)

MariaLStefan (changes requested):
Nicely done 👌
Just some optimizations

packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
```diff
@@ -67,9 +66,21 @@ export class NavDrawer implements ComponentInterface {
...
+   */
+  @Prop() isMobile?: boolean = false;
+
+  // set the variant based on the isMobile prop
+  @Watch('isMobile')
+  handleIsMobileChange() {
+    if (this.isMobile) {
+      this.variant = 'modal';
+    }
+  }
```
janivo, line 83: What's the need for changing the `variant` property? As far as I can see the `variant` prop only applies different classes in the `render` method. The same can be achieved by also adding the classes when the `isMobile` class is set. 

The fact that the `isMobile` makes the `variant` prop obsolete when it is set to `true` might also be an indicator that the mobile view is just another `variant`. What do you think about adding `mobile` as another variant?

packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
```diff
@@ -79,17 +90,28 @@ export class NavDrawer implements ComponentInterface {
...
 
+  private modifyMobileItems() {
+    if (!this.isMobile) return;
+
+    const navItems = this.el.querySelectorAll('ino-nav-item');
+
+    navItems.forEach((item) => {
+      item.classList.add('mobile-nav-item');
+    });
+  }
```
janivo, line 113: When transitioning from the mobile view to the desktop view, we also need to remove these classes again. That's currently not done, right?

It's probably a good idea to separate all the logic thats mobile specific into two methods like `activateMobileMode`and `deactivateMobileMode` or something like that.

janivo (changes requested):

janivo (approved):

PR #1041, opened 2023-08-07
feat(elements|ino-label): add for attribute to link from controls

Sl1nd:
Closes #...

## Proposed Changes

- add 'for' attribute to ino-label component
- inp-input: fix orphaned label error 
- ino-textarea: fix orphaned label error

### DIFF
--- a/packages/storybook/elements-stencil-docs.json
+++ b/packages/storybook/elements-stencil-docs.json
@@ -4279,6 +4279,22 @@
           "required": false
         },
         {
+          "name": "for",
+          "type": "string",
+          "mutable": false,
+          "attr": "for",
+          "reflectToAttr": false,
+          "docs": "Id of the associated form control",
+          "docsTags": [],
+          "values": [
+            {
+              "type": "string"
+            }
+          ],
+          "optional": false,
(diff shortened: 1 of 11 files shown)


packages/storybook/elements-stencil-docs.json
```diff
@@ -4278,6 +4278,22 @@
...
+          "reflectToAttr": false,
+          "docs": "Id of the associated form control",
+          "docsTags": [],
+          "values": [
+            {
+              "type": "string"
+            }
+          ],
+          "optional": false,
+          "required": false
```
BingeCode, line 4295: Just curious about this config: Doesn't `"optional": false` imply that it is required? Or how is optional and required to be understood in this context?

BingeCode (approved):

packages/storybook/elements-stencil-docs.json
janivo: That's actually not quite clear to me either. The generation of this file is done by stencil so maybe the documentation has some infos: https://stenciljs.com/docs/doc-generation

janivo (approved):

PR #1068, opened 2023-09-29
style(elements|ino-select): use new design

### LINKED ISSUE #1054: ino-select: update design
Our designers have come up with a new design for the ino-select component. Now it is time to implement it. 

A few guidelines:

- [ ]  the consumer API of the component should not be changed
- [ ] the input field of the ino-select and the ino-input should look the same (see #774)
- [ ] the options of the ino-select and the ino-autocomplete should look the same
- [ ] where possible the mixins of MDC should be used

MariaLStefan:
Closes #1054 

## Proposed Changes

- update design of `ino-select`
- add `_option-base.scss` with mixins to guaranty the same new design look on `ino-autocomplete`
- use `_option-base.scss` on `ino-option` & `ino-autocomplete`

## Preview
`ino-select`:

https://github.com/inovex/elements/assets/103122411/b5271432-0a0f-4b90-a274-9473fde04ce0

`ino-autocomplete`:

https://github.com/inovex/elements/assets/103122411/684d730d-ee33-48f4-8f61-bfb0d729f718

## Note
Keyboard navigation on `ino-autocomplete` shows the selected color state, because `ino-autocomplete` selects the item the user keyboard navigated to.
If the user hovers, it shows the deselected color state because no selection happened.


packages/elements/src/components/ino-autocomplete/ino-autocomplete.scss
```diff
@@ -1,41 +1,69 @@
...
 
+$option-selected-background-color: #{theme.$p-1};
+$option-selected-background-color-hover: #{theme.$p-2};
+$option-selected-background-color-focus: #{theme.$p-3};
+$option-selected-background-color-active: #{theme.$p-3};
+
+$option-deselected-background-color: #{theme.$white};
+$option-deselected-background-color-hover: #{theme.$n-1};
+$option-deselected-background-color-focus: #{theme.$n-2};
+$option-deselected-background-color-active: #{theme.$n-3};
```
janivo, line 16: Aren't these colors the same for the `ino-option`? If that's the case, we should directly define them in our `_option-base.scss` so we don't have to define the same colors values in our autocomplete and select component

janivo: I just saw that the colors of the `ino-option` are configurable. In that case, I would define constants in the `_option-base.scss` and use them as the fallback values for our CSS Variables.

Like this:
```scss
// in option-base.scss
$SELECTED_BG_COLOR: #{theme.$p-1}

// in ino-option.scss
@use '../base/option-base' as base;

  --option-selected-background-color: var(
    --ino-option-selected-background-color,
    base.$SELECTED_BG_COLOR
  );
```

That way, we've got all our color states (selected, unselected, active, ...) in the `_option-base.scss` and don't have to repeat them in our `ino-select` or `ino-autocomplete`

janivo (changes requested):
Looks great 👍🏼 
Just one thing to prevent duplication of our color states

janivo (approved):

PR #1072, opened 2023-10-04
chore: setup patterns package

### LINKED ISSUE #1065: Setup patterns package
In order to create and maintain our patterns, we would like to have a dedicated package. This package should be used as our "Playground" to implement all the UI patterns we have found and also should be responsible for creating importable react components for our landing page.

This package should be a simple react application with Vite and tailwindcss integrated. The application should contains our patterns which are defined below. The result of the build should include our plain `.html` files as well as our `.html` that are wrapped inside react components. 

A single pattern inside this package should contain the following:
- A `.html` file as our implementation of the component
- All styles should be defined via tailwindcss (no external css files)
- A react component that wraps the `.html` file that is exported and can be used within our landing page (see https://stackoverflow.com/q/40108843)

### Visualization

Image

> Why Tailwind?

Because it's easier to read HTML-Code without any additional CSS classes in an external File. Also, the use of TailwindCSS allows the user to copy only the HTML-Code and paste it into their project.

TobiasHeimGalindo:
Closes #1065


packages/ui-patterns/src/App.css
janivo, line 1: can be removed

packages/ui-patterns/src/App.tsx
```diff
@@ -0,0 +1,12 @@
+import '@inovex.de/elements-react';
+import MyPattern from './patterns/my-pattern-a/MyPatternA';
+
+function App() {
+  return (
+    <div className="App">
```
janivo, line 6: `className` can be removed

packages/ui-patterns/README.md
janivo, line 1: Could you write some lines about this package?

- What is its motivation?
- How does this package work internally (building a lib, copying html files to landingpage, ...)?
- How do I add another pattern?

janivo (changes requested):

PR #1077, opened 2023-10-17
feat(elements|ino-avatar): add loading animation

### LINKED ISSUE #1029: ino-avatar: add loading animation when image is being fetched
The animation should be designed in collaboration with our designers.
The animation could be the border spinning or just a simple pulsing animation.

Acceptance criteria:
- When providing an image via the `src` property, the avatar component automatically displays its fetch animation until the fetch has been completed (see https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event)
- A property exists to control the loading animation (independent of the above-mentioned automatic loading state; the automatic loading behavior should be overwritten if the consumer provided either `true` or `false`)
- A story exists, which shows an avatar and a checkbox. If the checkbox is checked, the avatar should start the loading animation and if the checkbox is unchecked, the animation should be stopped

TobiasHeimGalindo:
Closes #1029 

Preview:

https://github.com/inovex/elements/assets/81302108/240dd063-89ef-4441-9550-36ff34b283e7


packages/elements/src/components/ino-avatar/ino-avatar.tsx
```diff
@@ -53,6 +61,27 @@ export class Avatar implements ComponentInterface {
    * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role.
    */
   @Prop() a11yLabel?: string = '';
+  /**
+   * Overrides the avatar's loading animation behavior.
+   * When set to true, the loading animation is displayed indefinitely.
+   * When set to false, the avatar will not show any loading animations.
+   * 
+   * By not using this property or setting it to undefined, the loading animation will be shown only while the image is being loaded.
```
janivo, line 69:
```suggestion
   * By default, the loading animation will be shown only while the image is being fetched.
```

packages/elements/src/components/ino-avatar/ino-avatar.tsx
```diff
@@ -53,6 +61,27 @@ export class Avatar implements ComponentInterface {
...
+   */
+  @Prop() overrideLoading?: boolean = false;
```
janivo, line 71: I would name it `isLoading` and rename our internally used variable (e.g. `imgIsFetching` or `_isLoading`)

packages/elements/src/components/ino-avatar/ino-avatar.scss
```diff
@@ -122,10 +132,49 @@
...
+      animation: spinner 1s linear infinite;
+    }
+
+    @keyframes spinner {
+      0% {
+        transform: rotate(0deg);
+      }
+      100% {
+        transform: rotate(360deg);
+      }
```
janivo, line 175: I feel like the animation might be a bit too intrusive. Imagine having a list of employees that's shown and all avatars are fetching at the same time. This might be a little bit too much for the user with our current animation.

What do you think about some skeleton-like animation like [this one](https://codepen.io/maoberlehner/pen/bQGZYB)?

janivo (changes requested):
Nice job 👍🏼 
But I think there might be some room for improvement to the loading animation as it might be too intrusive for the user

packages/elements/src/components/ino-avatar/ino-avatar.scss
TobiasHeimGalindo: Good idea. I added a modified skeleton-animation for the ino-avatar, check out the video in the PR description.

janivo (approved):
I just went through our components and did not see any occurrences of the `is` prefix for boolean props. So I renamed `isLoading` to `loading` again. My bad here 😅 

Nevertheless, the animation looks sick. Really nice 😍

PR #1080, opened 2023-10-23
feat(landingpage): add patterns preview-code-box

### LINKED ISSUE #1076: Implement preview-code-box for our patterns section
### Feature Request

In preparation for our patterns section, we need a way to present our patterns to our consumers. The consumer should be able to see how the pattern looks when rendered and also how he/she can achieve this pattern via code.

In order to do so, we want to implement ...
-  a box with a header
- a toggle button to switch between preview and code
- the box itself where ...
  - the pattern is rendered when the toggle is switched to preview
  - the code is rendered when the toggle is switched to code
- the code snippet should also provide ...
  - a copy button to copy the source code by clicking
  - an HTML syntax highlighting of the shown code snippet

**Have a look at the design in Figma.**

### Additional context

- The framework selection can safely be ignored as we only offer our code in HTML
- I think we've already some kind of syntax highlighter in our guides, try to use that before integrating something else
- Use the `ino-segment-group` or the `ino-switch` (icon variant) to switch between preview and code mode

Image

TobiasHeimGalindo:
Closes #1076 

Notes: once deployed, the patterns page will be filtered out of our routes(non-existant in our navbar and footer) and if searching the url /library/patterns it will show our Custom404 screen to redirect to our Home Page

Preview:

https://github.com/inovex/elements/assets/81302108/c803c6e0-3be8-4231-baa8-a1b01ce462ba


packages/landingpage/pages/[lang]/library/patterns/preview-box.tsx
Comment (janivo) [inline packages/landingpage/pages/[lang]/library/patterns/preview-box.tsx:1]: That's a rather big component. What do you think about splitting it up into many components (could even be in this file)?

These could be our Components (all located in the `preview-box.tsx` file):
- CopyButton
- ViewModeSelection
-  PreviewBox

That way we don't have these many hooks that actually only do one thing (e.g. `handleIconClick`, `copyIcon` and `visiblePopover` only really concern the copy button)

packages/landingpage/pages/[lang]/library/patterns/preview-box.tsx
```diff
@@ -0,0 +1,156 @@
...
+        {selectedValue === 'Preview' && (
+          <div>{previewComponent}</div>
+        )}
+        {selectedValue === 'Code' && (
+          <div
+            className={styles.code}
+            dangerouslySetInnerHTML={{ __html: highlightedCode }}
+          ></div>
+        )}
+      </div>
```
Comment (janivo) [inline packages/landingpage/pages/[lang]/library/patterns/preview-box.tsx:153]: It feels a little weird to have the Box jumping at its height. Let's talk about this in our meeting tomorrow.

packages/landingpage/pages/[lang]/library/patterns/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    highlightedCodes[label] = await highlightCodeWithShiki(code);
+  }
+  return {
+    props: {
+      ...languageProperties,
+      highlightedCodes,
+      codeStrings,
+    },
+  };
+};
```
Comment (janivo) [inline packages/landingpage/pages/[lang]/library/patterns/index.tsx:115]: Very nice solution!

janivo (changes requested):
Very nicely done! It's really fun clicking around because of those little gimmicks you've implemented 💯

janivo (approved):

PR #1096, opened 2023-11-08
refactor(elements|ino-accordion): accessible ino-accordion

### LINKED ISSUE #1003: Accessible InoAccordion
### Feature Request

Accessible InoAccordeon according to the W3C-Aria pattern https://www.w3.org/WAI/ARIA/apg/patterns/accordion/

According to the pattern some findings are:

* accordion header is not container in an element with role button
* 'aria-expanded' property is missing
* 'aria-controls' property is missing 
* accordeon content should be marked properly
 
* The title of each accordion header is contained in an element with role [button](https://w3c.github.io/aria/#button).
* Each accordion header button is wrapped in an element with role [heading](https://w3c.github.io/aria/#heading) that has a value set for [aria-level](https://w3c.github.io/aria/#aria-level) that is appropriate for the information architecture of the page.
   * If the native host language has an element with an implicit heading and aria-level, such as an HTML heading tag, a native host language element may be used.
   * The button element is the only element inside the heading element. That is, if there are other visually persistent elements, they are not included inside the heading element.
* If the accordion panel associated with an accordion header is visible, the header button element has [aria-expanded](https://w3c.github.io/aria/#aria-expanded) set to true. If the panel is not visible, [aria-expanded](https://w3c.github.io/aria/#aria-expanded) is set to false.
* The accordion header button element has [aria-controls](https://w3c.github.io/aria/#aria-controls) set to the ID of the element containing the accordion panel content.
* If the accordion panel associated with an accordion header is visible, and if the accordion does not permit the panel to be collapsed, the header button element has [aria-disabled](https://w3c.github.io/aria/#aria-disabled) set to true.
* Optionally, each element that serves as a container for panel content has role [region](https://w3c.github.io/aria/#region) and [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) with a value that refers to the button that controls display of the panel.
    * Avoid using the region role in circumstances that create landmark region proliferation, e.g., in an accordion that contains more than approximately 6 panels that can be expanded at the same time.
    * Role region is especially helpful to the perception of structure by screen reader users when panels contain heading elements or a nested accordion.

TobiasHeimGalindo:
Closes #1003


packages/elements/src/components/ino-accordion/ino-accordion.tsx
```diff
@@ -19,15 +20,17 @@ import classNames from 'classnames';
 })
 export class Accordion implements ComponentInterface {
   @Element() el: HTMLInoAccordionElement;
+  @State() contentId: string;
```
janivo, line 23: Instead of introducing a state variable that triggers an extra rerender after it has initially been loaded, I would suggest just using a private variable that is initialized on creation. Like this:

```ts
private readonly contentId = `ino-accordion-content-${this.el.id}`
```

This variable can then be used in the `render` method without triggering an extra rendering process.

janivo (changes requested):

janivo (approved):

PR #1097, opened 2023-11-09
refactor(elements|ino-dialog): accessible ino-dialog

### LINKED ISSUE #1006: Accessible InoDialog
### Feature Request

Accessible dialog according to the [WAI-ARIA Dialog Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)  

According to the pattern some findings are:

* the keyboard interaction pattern the ESC key should close the dialog. Currently this behavior could be disabled via `dismissible` api which leads to prevent keyboard users from using the dialog. 
* [aria-modal](https://w3c.github.io/aria/#aria-modal) property should only be set if user may close the dialog by clicking outside.
* role alertdialog should only be used to divert users' attention to a brief, important message. Could not be used statically

* The element that serves as the dialog container has a role of [dialog](https://w3c.github.io/aria/#dialog).
* All elements required to operate the dialog are descendants of the element that has role dialog.
* The dialog container element has [aria-modal](https://w3c.github.io/aria/#aria-modal) set to true.
* The dialog has either:
* A value set for the [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) property that refers to a visible dialog title.
* A label specified by [aria-label](https://w3c.github.io/aria/#aria-label).
* Optionally, the [aria-describedby](https://w3c.github.io/aria/#aria-describedby) property is set on the element with the dialog role to indicate which element or elements in the dialog contain content that describes the primary purpose or message of the dialog. 
Specifying descriptive elements enables screen readers to announce the description along with the dialog title and initially focused element when the dialog opens, which is typically helpful only when the descriptive content is simple and can easily be understood without structural information. It is advisable to omit specifying aria-describedby if the dialog content includes semantic structures, such as lists, tables, or multiple paragraphs, that need to be perceived in order to easily understand the content, i.e., if the content would be difficult to understand when announced as a single unbroken string.

TobiasHeimGalindo:
Closes #1006

### DIFF
--- a/packages/elements/src/components/ino-dialog/ino-dialog.tsx
+++ b/packages/elements/src/components/ino-dialog/ino-dialog.tsx
@@ -46,7 +46,7 @@ export class Dialog implements ComponentInterface {
   @Prop() fullwidth?: boolean;
 
   /**
-   * Close the dialog on pressing the ESC key or clicking outside of the dialog.
+   * Close the dialog by clicking outside of the dialog.
    */
   @Prop() dismissible?: boolean;
 
@@ -54,6 +54,12 @@ export class Dialog implements ComponentInterface {
    * Opens the dialog if set to true
    */
   @Prop() open = false;
+
+  /**
+   * The role of the dialog. Can be either 'dialog' or 'alertdialog'.
+   * The 'alertdialog' role should be used for important alerts and error messages.
+   */
+  @Prop() dialogRole?: 'dialog' | 'alertdialog' = 'dialog';
 
   /**
    * Adds a headline to the `ino-dialog`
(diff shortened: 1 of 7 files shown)


packages/elements/src/components/ino-dialog/ino-dialog.tsx
```diff
@@ -87,7 +93,7 @@ export class Dialog implements ComponentInterface {
 
   @Listen('keyup', { target: 'body' })
   handleKeyUp(event: KeyboardEvent) {
-    if (event.key === 'Escape' && this.open && this.dismissible) {
+    if (event.key === 'Escape' && this.open) {
```
janivo, line 96: Why did you change this condition? Should a dialog always be closeable by pressing escape? 🤔

janivo (approved):

packages/elements/src/components/ino-dialog/ino-dialog.tsx
TobiasHeimGalindo: Yes, based on the findings of #1006 the `ESC` key should always be able to close the dialog - this is to ensure that keyboard users can easily close the dialog, especially for those who rely on screen readers or cannot use a mouse.

janivo (approved):

PR #1098, opened 2023-11-10
refactor(ino-tab-bar): accessible ino-tab-bar

### LINKED ISSUE #1002: Accessible InoTabBar
### Feature Request
Implement InoTabBar according to https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

TobiasHeimGalindo:
Closes #1002 

### Note on Accessibility Implementations for Tab Panels

Our `ino-tab` and `ino-tab-bar` components focus on providing the tab and the tablist functionality. We do not control or provide the tab panels, as these are part of the user's content implementation. To support accessibility, I added a `panelId` property to the `ino-tab` component. This enables users to link each tab with its corresponding content panel for accessibility purposes, following WAI-ARIA practices. However, the actual implementation of tab panels and their accessibility attributes (`role="tabpanel"`, `aria-labelledby`, etc.) is under the user's control.

### DIFF
--- a/packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
+++ b/packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
@@ -30,13 +30,6 @@ export class TabBar implements ComponentInterface {
    */
   @Prop() activeTab?: number = 0;
 
-  @Watch('activeTab')
-  activeTabChangedWatcher(newTabIndex: number) {
-    if (this.mdcInstance) {
-      this.mdcInstance.activateTab(newTabIndex);
-    }
-  }
-
   /**
    * Autofocus of tab on activation.
    */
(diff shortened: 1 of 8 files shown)


packages/elements/src/components/ino-tab-bar/ino-tab-bar.tsx
```diff
@@ -48,10 +41,34 @@ export class TabBar implements ComponentInterface {
...
+    tabs.forEach((tab, index) => {
+      const isSelected = index === activeTabIndex;
+      const panelId = (tab as HTMLInoTabElement).panelId;
+
+      const buttonElement = tab.querySelector('button');
+
+      buttonElement.setAttribute('aria-selected', isSelected.toString());
+      if (panelId) {
+        buttonElement.setAttribute('aria-controls', panelId);
+      }
```
janivo, line 63: These lines assume the internal structure of the `<ino-tab>` component and therefore have too much knowledge. As soon as we do some refactoring of our tab component, we don't really realize that we might be breaking things.

That's why I would propose to implement these lines as properties of the `<ino-tab>` component. They could be named like our other accessibility props (e.g. `a11ySelected` and `a11yControls`) and just set these values to the internal `button` element. We also should write a disclaimer in the comments that these properties are managed by the `ino-tab-bar` component.

This way, the `ino-tab-bar` uses only the public API of our `ino-tab`.

janivo (changes requested):

janivo (changes requested):

Issue #1058, opened 2023-09-12
Evaluate components conflicting with dark-mode


janivo:
In order to gain some insights on how much time the implementation of the dark mode is going to take, we need a list of components that need to have a tweak to their design in order to be compatible with the dark mode.

Check for every component whether it is compatible with the dark mode. If you think it is not a big effort to make the respective component compatible with dark mode (e.g. changing `background-color` from `white` to `transparent`), feel free. Otherwise, if a component is incompatible and it's a bigger effort to fix it, add it to the list below.

There is an easy way to check it in the storybook:
![Image](https://github.com/inovex/elements/assets/22963121/7bd9dda1-4ce7-4d7d-a2a1-9acad405b0c3)

### List of dark-mode incompatible components:
- ...

PR #1110, opened 2023-11-27
feat(elements): add light- & dark-mode configuration

### LINKED ISSUE #1057: Enable configuration of light- and dark-mode
One of our first steps for introducing dark mode to the elements would be to create a global way to configure light or dark mode for all of our components.

Have a look at our [configuration file](https://github.com/inovex/elements/blob/d370dc254d583753971017536461566b9a65cd7a/packages/elements/src/components/config.ts) on how to add such an option.

TobiasHeimGalindo:
Closes #1057

**Theme switch example** _(highlighted green only for the accordion--expanded component currently)_:

https://github.com/inovex/elements/assets/81302108/56816cb5-db5c-4910-8171-c7dfaa5b49bc

### Notes: 

- Defined new primary/neutral/alert colors in our global `styles.scss`
- Enable our inovex Elements components to be theme aware by using CSS variables (` var(--inovex-elements-p-1);` instead of `theme.$p-1;`).
- Switch the css class in the landingpage and storybook to `.inovex-elements-theme` & `.inovex-elements-theme-dark`
- Disabled all features for production using `isInDevEnvironment` flag

## TODOS for 100% light- & dark-mode support

1. Add a dark theme color pallete covering all our neutral, primary and alert colors to our global styles.scss in the elements package

2. Stencil Components: Switch to CSS variables for light- & dark theme support _(enables theme-awareness at runtime)_

3. Design two new Ino-Icons for theme switch on landingpage: sun & moon icon _(light & dark ino-switch toggle)_

### DIFF
--- a/packages/elements/src/components/ino-accordion/ino-accordion.scss
+++ b/packages/elements/src/components/ino-accordion/ino-accordion.scss
@@ -40,7 +40,7 @@ $animation-time: 200ms;
   }
 
   &--expanded {
-    background: theme.$p-1;
+    background: var(--inovex-elements-p-1); //Example for CSS variable usage
   }
 
   .ino-accordion__content-wrapper {
--- a/packages/elements/src/components/ino-accordion/ino-accordion.scss
+++ b/packages/elements/src/components/ino-accordion/ino-accordion.scss
@@ -6,6 +6,11 @@ $animation-time: 200ms;
 .ino-accordion {
   &:hover {
     background: theme.$p-1;
+    /*
+    TODO: Switch to CSS variables for light- & dark theme support.
+    Replace the above line with the following once the global styles.scss file has been updated with a dark-theme import:
+    background: var(--inovex-elements-p-1);
+    */
   }
   border-radius: 24px;
 
--- a/packages/elements/src/global/styles.scss
+++ b/packages/elements/src/global/styles.scss
@@ -144,13 +144,7 @@
 
 // Dark Theme Variables
 .inovex-elements-theme-dark {
-  --inovex-elements-white: #{dark-theme.$white};
-  --inovex-elements-black: #{dark-theme.$black};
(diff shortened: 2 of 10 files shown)


janivo:
TODO: check MDC mixin compatibility

packages/elements/src/components/ino-accordion/ino-accordion.scss
```diff
@@ -35,7 +40,7 @@ $animation-time: 200ms;
   }
 
   &--expanded {
-    background: theme.$p-1;
+    background: var(--inovex-elements-p-1); //Example for CSS variable usage
```
MariaLStefan, line 43: This is a great start for switching between light and dark theme!
The approach of using a second theme-color-pallet with the same names as the given light-theme is great, because we won't need to introduce more variables.

But i have the concern, that we won't have a dark theme-color-pallet which will perfectly match with the light-theme colors. 

Because some components might work which their light-theme colors in dark mode but some might need bigger adjustments which won't perfectly fit across all other components. 

If we won't find a fitting dark-theme-color-pallet I would suggest to store the component colors as variable (e.g. `$ino-accordion-bg-color`) which will choose, according to the selected theme, what color should be displayed. This will allow a fine granular adjustment of all colors.

Example: 
**`ino-accordion-colors.scss`**(holds only accordion colors) or **`ino-components-colors.scss`**(holds colors of every component)
````
.inovex-elements-theme-light {
 $ino-accordion-bg-color: theme.$p-2;
 $ino-accordion-icon-color: theme.$p-5;
 $ino-accordion-text-color: theme.$n-11;
 ...
}

.inovex-elements-theme-dark {
 $ino-accordion-bg-color: theme.$n-8;
 $ino-accordion-icon-color: theme.$p-5;
 $ino-accordion-text-color: theme.$white;
 ...
}
````
But with smaller deviation we won't need this procedure.

Waiting for #1058 and dark-theme-pallet

PR #1147, opened 2023-12-22
feat(elements): dialog close icon

### LINKED ISSUE #1069: ino-dialog: Add option to display cancel button
Usually dialogs have a cancel button in the top right or left corner of the dialog.
This should be configurable by a flag when using the dialog.

Currently:

Image

iandi09:
Closes #1069

- added an optional property to display a close icon in the top right corner of the dialog


TobiasHeimGalindo (approved):

packages/elements/src/components/ino-dialog/ino-dialog.scss
MariaLStefan, line 1: We try to avoid access nesting because it slows down the debugging process. 
Example: Searching class names. 
Your browser dev tools will show the full class name, by copy-pasting it into your IDEs search bar you will find it pretty quickly. By breaking the classnames down into the scss selectors the search process will take a hit, because you have to dig deeper into the encapsulations. 
I understand that your approach is easily read, can be written pretty fast and looks sleek, but this repo is maintain through a lot of people through out the years, which will be debugging and housekeeping a lot of the time. Because of that we try to minimize the time consumed by this by thoroughly naming the selectors in our stylesheets.

MariaLStefan (changes requested):
Nicely done and works like a charm 💪

To finish this issue, I would suggest adding a story to showcase the new prop and for a polished look it would be great to add a hover and clicked state to the closing icon.

packages/elements/src/components/ino-dialog/ino-dialog.scss
```diff
@@ -36,55 +36,63 @@ $material-dialog-box-shadow: 0px 0px 22px rgba(93, 91, 105, 0.35);
...
-  .mdc-dialog .mdc-dialog__surface header h1 {
-    margin: 0 0 0;
-    font-weight: 400;
-  }
 
-  .mdc-dialog .mdc-dialog__surface section {
-    margin: 38px 0 0;
-  }
+    &__container {
+      height: $height;
```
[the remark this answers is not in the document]
iandi09: Oh ok, I wasn't aware of that problem. But as far as i understand the referenced article, it shouldn't be a problem in this scenario because the sass variable is not used in any css variable (the exposed variable is --ino-dialog-height). Or am I wrong? I've tested it as well and everything works as expected.

packages/elements/src/components/ino-dialog/ino-dialog.scss
iandi09: In my opinion we get more benefits from nesting because it's much easier to understand the structure of the css and the specificity of the selectors. And it's easier to identify unnecessary styles (for example due to the duplication of several selectors).

In terms of debugging I would say it's still comfortable with nesting because we see the name of the component in the css class and can directly go to the related scss file. And there it should be easy to find the selector we are looking for in case the file is well structured.

But since I'm new to the project I respect your decision on this topic and if you want I can remove the nesting again. :)

MariaLStefan (approved):

PR #1149, opened 2023-12-22
fix(elements-angular): make ng-add work with newer angular versions

janivo:


BenPag (changes requested):
How do you tested it? I tried it with yalc but i got some compilation errors.

image

But the ng-add command seems to be working

janivo:
> How do you tested it? I tried it with yalc but i got some compilation errors.
> 
> image
> 
> But the ng-add command seems to be working

I tested it by building it locally, moving it to another independent angular application and running `ng add <path to build tar>`. Afterwards, to make it runnable you've got to replace the `inovex.de/elements` version in the `node_modules` of the independent by hand with a build of the same branch. Kinda complicated but I was able to test it successfully 😅

packages/elements-angular/src/schematics/utils/ast.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export function addElementsModuleImportToNgModule(host: Tree, modulePath: string): void {
+  const recorder = host.beginUpdate(modulePath);
+  const moduleSource = getSourceFile(host, modulePath) as any;
+
+  const ionicModuleChange = insertImport(moduleSource, modulePath, 'InoElementsModule', '@inovex.de/elements-angular');
+
+  applyToUpdateRecorder(recorder, [ionicModuleChange]);
+
+  const metadataChange = addSymbolToNgModuleMetadata(moduleSource, modulePath, 'imports', 'InoElementsModule.forRoot()');
```
BenPag, line 31: In my example project only the InoElementsModule will be added to the import list. The necessary `InoElementsModule.forRoot()` statement was not added in the @NgModule imports array.

BenPag (approved):

PR #1152, opened 2023-12-29
feat(storybook) added controls for CSS variables

### LINKED ISSUE #453: Stories for Custom-Properties (CSS-Vars)

iandi09:
Closes #453

- added css variable support to story template generator
- extracted css props for the existing stories


janivo:
It would be cooler to extract this CSS prop from the `stenci-docs.json` file, so we can automatically generate the respective controls.

PR #1182, opened 2024-01-18
feat(elements|ino-markdown-editor): allow custom font-sizes via CSS variables

### LINKED ISSUE #1155: ino-markdown-editor: Make font-sizes configurable
Our `ino-markdown-editor` works with fixed font-sizes for all kinds of text (paragraphs, headlines, ...). These sizes won't be fitting for all of our consumers, so we need to make them adjustable. Either by using CSS-Variables (preferred) or by properties.

TobiasHeimGalindo:
Closes #1155


MariaLStefan (approved):

packages/elements/src/components/ino-markdown-editor/ino-markdown-editor.scss
```diff
@@ -149,6 +153,15 @@ ino-markdown-editor {
...
     color: rgba(0, 0, 0, 0.87);
+    font-size: var(--ino-markdown-editor-paragraph-font-size, 1rem);
+
+    h1 {
+      font-size: var(--ino-markdown-editor-h1-font-size, 2rem);
+    }
+
+    h2 {
+      font-size: var(--ino-markdown-editor-h2-font-size, 1.75rem);
+    }
```
janivo, line 164: I've got an idea that would make it a little more consumer-friendly: Instead of assigning each text type (h1, h2, p) its own value, how about offering a variable that calculates the other values based on it? This has the advantage that I only have to enter one value as a consumer (e.g. `12px`) and the other values can be calculated automatically (e.g. `h1 = 12px * 1.5`, `h2 = 12px * 1.25`, ...). Otherwise, as a consumer, I always have to specify these 3 variables at the same time, as the ratio between the font sizes is would no longer be correct.

janivo (changes requested):
Just an idea. Would love to hear your thoughts on it.

packages/elements/src/components/ino-markdown-editor/ino-markdown-editor.scss
TobiasHeimGalindo: Good idea - I agree, by dynamically scaling with `calc()`, the user only needs to specify one base font size without much thought, while maintaining the correct relative ratios of `h1` and `h2`.

janivo (approved):

PR #1183, opened 2024-01-18
fix(landingpage): routing of shared storybook links doesn't work

### LINKED ISSUE #1156: Storybook<>Landingpage Links dont work
## Steps to reproduce
1. Go to Storybook in our Landing page
2. Open any page that is not the welcome page
3. Copy the link to your current page
4. Go to a new tab and paste the link
5. See that it always opens the welcome page even though the previously copied page should be opened

## GIF
Image

## Possible solutions
We've implemented our [own storybook plugin](https://github.com/inovex/elements/blob/master/packages/storybook/addons/post-current-story.tsx) that uses the `<iframe>` post message API to send the currently opened storybook URL to our landing page and the landing page should then reflect the link in the URL in the browser. That seems to work just fine. 

When opening such a reflected URL, the landing page should be able to parse it to an actual storybook URL that can be passed to the `<iframe>` element. This part seems not to be working properly. This feature is implemented [here](https://github.com/inovex/elements/blob/master/packages/landingpage/utils/hooks/useStorybookUrl.ts).

TobiasHeimGalindo:
Closes #1156 

## Notes:

Ever since I added the [selectedVersion dropdown](https://github.com/inovex/elements/commit/9028657c6841848437d6acd2332ff50cc9d632c0), the iframe will default to the welcome page (`https://elements.inovex.de/version/${selectedVersion}/?path=/story/docs-welcome--page`)

to prevent this defaulting on each refresh the correct path is now extracted from the initial URL and constructed in the iFrameUrl useMemo. The `useStorybookUrl.ts` and `useStorybookUrlSyncer.ts `were okay


MariaLStefan:
I still get this error when i want to access a page via url paste. I fixed this in https://github.com/inovex/elements/pull/1131 ...
Did something break, or am I the only one who gets this error?

> Error: The provided `href` (/[lang]/library/en) value is missing query values (lang) to be interpolated properly. Read more: https://nextjs.org/docs/messages/href-interpolation-failed

janivo (approved):

PR #1185, opened 2024-01-18
chore: enhance release process

### LINKED ISSUE #1179: Enhance Release Process and Address Recent Issues
## Description
In light of recent changes in our release process, transitioning from Lerna to full NX releases, we encountered several issues during the initial release attempt. This task encompasses the resolution of these issues and further improvements to our release workflow.

#### Pipeline Push Permissions and Dry Run Release
Investigate and address the pipeline push permission issues that occurred during the first release attempt. Additionally, ensure the setup still allows for triggering a dry run release for enhanced testing and validation. Have a look at [this workflow](https://github.com/inovex/elements/actions/runs/7224434985) for more information.

#### Local Push Configuration for Release Notes
Confirm and configure push settings to enable the correct pushing of release notes to GitHub. This is crucial for maintaining accurate release documentation.

#### Unicorn Failure on Master Branch
Investigate and troubleshoot the Unicorn failure on the master branch during the initial release. Implement corrective measures to prevent recurring failures.

#### Preleases should be derived correctly
With our current pre-release process, no pre-release tag will be pushed to the master branch. Therefore, our release tool cannot correctly derive the current prerelease version.

#### Prerelease Tag Exclusion from `latest`
Implement a solution to prevent prereleases from receiving the `latest` tag. Prereleases should remain associated with the last successful major/minor/patch release. 

#### Bump internally used version
When releasing a new version, our internally used dependencies don't get bumped like in lerna before. For example, the `@inovex.de/elements-react` package depends on `@inovex.de/elements` which will not get bumped in case of a new major release.

## Acceptance Criteria:

- [ ] Pipeline push issues are resolved, allowing for successful releases
- [ ] Local push settings are confirmed and configured appropriately to ensure accurate and timely release note documentation.
- [ ] The recurring Unicorn failure on the master branch is investigated, and corrective actions are taken to prevent future failures. It is also possible to create a new tasks with the findings.
- [ ] Prereleases are successfully derived with every merge into the master branch
- [ ] Prereleases are excluded from receiving the `latest` tag, maintaining their association with the last successful major/minor/patch release.
- [ ] Bump internally used elements version to new release automaitcally
- [ ] Ensure that the user has the respective permissions to push to master and also has the permission to publish a package to NPM BEFORE running the actual push/publish

BenPag:
Closes #1179

## Proposed Changes

- use nodejs script to execute release
- use new `nx release` command
- enable dry-run mode for workdlow
- add check if user is logged into npm
- provide GitHub release (not for prerelease)
- automate prerelease (canary)


scripts/release.js
```diff
@@ -0,0 +1,48 @@
+const shell = require('shelljs');
+const semver = require('semver');
+const conventionalRecommendedBump = require('conventional-recommended-bump');
+
+async function getNextVersion() {
+  // get version from root package.json
+  // need to remove double quotes ("") from version string
+  const VERSION = shell.exec('npm pkg get version').trim().replace(/"/g, '');
```
janivo, line 8: Can't we get the version simply by doing it like this?
https://github.com/inovex/elements/blob/bfdfc0205fa3aed18f8667ca9f8cef3d5c10b834/scripts/bumprootversion.js#L3-L6

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+  const { releaseType } = await conventionalRecommendedBump({ preset: 'angular' })
+  return semver.inc(VERSION, releaseType);
```
janivo, line 10: Print out something like: Calculating next version ... 9.0.1 ✅

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const isDryRun = process.argv.some((a) => a.includes('dryRun=true'));
```
janivo, line 13: Print out: Dry run? ✅/❌

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+const isPreRelease = process.argv.some((a) => a.includes('canary=true'));
```
janivo, line 14: Print out: Prerelease? ✅/❌

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+
+const packagesToPublish = [
+  'packages/elements',
+  'packages/elements-angular',
+  'packages/elements-react',
+  'packages/elements-vue',
+];
+
+(async () => {
+  const currentBranch = shell.exec('git branch --show-current').trim();
```
janivo, line 25: Print out: Branch is master or prerelease? ✅/❌

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+  if (currentBranch !== 'master' && !isDryRun) {
+    shell.echo('Sorry, release is only on branch "master" allowed!');
+    shell.exit(1);
+  }
+
+  // check if user is logged in
+  if (shell.exec('npm whoami').code !== 0) shell.exit(1);
```
janivo, line 32: Print out: Logged into NPM? ✅/❌

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+
+  if (isPreRelease) {
+    shell.exec(`nx release version prerelease ${dryRunArg} --git-commit=false`);
```
janivo, line 35: Print out: Running pre release

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+  } else {
+    const version = await getNextVersion();
+    shell.exec(`nx release ${version} ${dryRunArg} --skip-publish`);
```
janivo, line 38: Print Out: Running Release (Version: 9.0.0, Dry run: ✅/❌)

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  if (shell.exec(`git push ${dryRunArg} --atomic --follow-tags`).code !== 0) shell.exit(1);
```
janivo, line 41: Print out: Pushing changes to $branch ... ✅/❌

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const npmTag = isPreRelease ? 'canary' : 'latest';
```
janivo, line 43: Print out: Calculating NPM tag ... canary / latest ✅

scripts/release.js
```diff
@@ -0,0 +1,48 @@
...
+  for (const npmPackage of packagesToPublish) {
+    const { code } = shell.exec(`npm publish -w ${npmPackage} --tag ${npmTag} ${dryRunArg}`);
+    if (code !== 0) shell.exit(1)
+  }
```
janivo, line 47: Print out: Publishing all Packages to NPM ...

- @inovex.de/elements ✅/❌
- @inovex.de/elements-react ✅/❌
...

janivo (changes requested):

scripts/release.js
janivo: also: Hint that you need to login into NPM with the respective premissions if it fails

scripts/release.js
janivo: Also on fail: Could not release package XYZ: <Error>

scripts/release.js
janivo: Also, on fail: Could not push changes to $Branch: <Error>

scripts/release.utils.js
janivo, line 1: I get why you are doing this (because the `release.js` file would get too large) but I think in this case, it hurts the readability since you constantly have to switch between those files to fully get all commands that are executed.

I would expect the `release.js` file to be structured like this:

```ts
// constants
const isDryRun = process.argv.some((a) => a.includes('dryRun=true'));
const isPreRelease = process.argv.some((a) => a.includes('canary=true'));
...

// function declarations
function getIconByExitCode(iconType) {
  switch (Number(iconType)) {
    case 0:
      return '❌';
    case 1:
      return '✅';
    case 2:
      return '☑️';
    default:
      return '⚠️';
  }
}
...

// main function that is executed
echo('-n', 'Getting latest release tag ... ');
```

scripts/release.js
janivo, line 1: Could you migrate this to a `.ts` file and call it with `ts-node` that is already integrated?

If it's not that easy, we can skip it

janivo (changes requested):
Some minor improvements. Also you might need to run the format fix

janivo (approved):

PR #1200, opened 2024-01-26
feat(elements/ino-snackbar): improve accessibility for all snackbar variants

### LINKED ISSUE #1199: [A11y]: InoSnackbar is still not fully accessible
### Bug Report

As snackbars optionally require an user action it should be possible to add role 'alertdialog' in addition to an accessible name expressing the need for action. In Addition currently its not possible to close the snackbar using "esc".

### Possible Solution

- Adding the possibility to configure the role of the snackbar ('alert', 'alertdialog')
- Close the  snackbar pressing "Esc"
- Hide icon from Screenreader as it is not accessible

Solution should be implemented referred to https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/examples/alertdialog/ as well as https://www.w3.org/WAI/ARIA/apg/patterns/alert/examples/alert/ to make all variants of ino-snackbar accessible.

Sl1nd:
Closes #1199 

## Proposed Changes

- Close on Esc
- Provide accessible name for alertdialog role as well as close icon button
- Configurable role property for snackbars that requires actions 
- Removed show close button on hover as it is not accessible via keyboard


packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
```diff
@@ -54,11 +55,25 @@ export class Snackbar implements ComponentInterface {
    */
   @Prop() timeout?: number = 5000;
 
+  /**
+   * The role of the dialog. Can be either 'dialog' or 'alertdialog'.
+   * The 'alertdialog' role should be used for important messages that requires user interaction.
+   */
+  @Prop() snackbarRole?: 'alert' | 'alertdialog' = 'alert';
```
janivo, line 62: As a default, can't we derive this state based on the existence of `actionText`?

e.g. Pseudocode
```
snackbarRole = this.snackbarRole ?? this.actionText ? 'alertdialog' : 'alert';
```

packages/elements/src/components/ino-snackbar/ino-snackbar.scss
```diff
@@ -138,6 +126,10 @@ ino-snackbar {
       color: #74758b;
     }
 
+    .ino-snackbar-action-container {
+      margin-top: 0.5rem;
```
janivo, line 130: IMO a smaller margin (e.g. `0.25rem`) would fit a lot better. What do you think? @BenPag 

image

```suggestion
      margin-top: 0.25rem;
```

janivo (changes requested):

packages/elements/src/components/ino-snackbar/ino-snackbar.scss
BenPag: I agree, a smaller margin fits better.

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
BenPag: I am not sure if this is suitable. In general I think we should rethink the default role too, becaue we use the snackbar for notifications that not needs an immediate attention from the user.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alertdialog_role);

> Warning: Because of its intrusive nature, the alert role must be used sparingly and only in situations where the user's immediate attention is required.

janivo: Lets implement it like this then:
```
if variant = 'error' || variant = 'warning' && hasAction -> 'alertDialog'
if variant = 'error' || variant = 'warning' -> 'alert'
else 'status'
```

packages/elements/src/components/ino-snackbar/ino-snackbar.scss
janivo: also fix alignment of X button

packages/elements/src/components/ino-snackbar/ino-snackbar.tsx
```diff
@@ -56,10 +56,10 @@ export class Snackbar implements ComponentInterface {
   @Prop() timeout?: number = 5000;
 
   /**
-   * The role of the dialog. Can be either 'dialog' or 'alertdialog'.
+   * The role of the dialog. Can be either 'status', 'alert' or 'alertdialog'.
    * The 'alertdialog' role should be used for important messages that requires user interaction.
    */
-  @Prop() snackbarRole?: 'alert' | 'alertdialog' = 'alert';
+  @Prop() snackbarRole?: 'status' | 'alert' | 'alertdialog';
```
janivo, line 62: Do we still need this property when we can just infer the correct role by ourselves? I would suggest removing it entirely.

janivo (changes requested):

janivo (approved):

Issue #865, opened 2023-01-18
elements-angular: Angular v15 support


Sl1nd:
### Feature Request

elements-angular currently implemented Angular v12 as peer dependency. 

### Description of the proposed solution

Upgrade elements package to Angular v15

Sl1nd:
I think the removal of 'ViewEngine'  in Angular v13 as well as the impact to our users should be documented expicitly in changelog.

PR #1215, opened 2024-01-31
fix(elements-angular): directives matching on multiple components

### LINKED ISSUE #1213: elements-angular: Directives registering twice
Resolve the "More than one custom value accessor matches form control with unspecified name attribute" Error

### Acceptance Criteria
- [ ] The error message "More than one custom value accessor matches form control with unspecified name attribute" is resolved.
- [ ] Directives are adjusted to avoid conflicting matches with form controls lacking a specified name attribute.
- [ ] Tests are implemented to prevent these errors in the future

janivo:
Closes #1213 

- We reworked our directives in #865 
- This lead to directives matching for multiple components, e.g. the example below was matched by our `NumericValueAccessorDirective` as well as `TextValueAccessorDirective` which led to runtime errors.
```html
<ino-input [formControl]="myFormControl" type="number"></ino-input>
```

### DIFF
--- a/packages/elements-angular/src/directives/control-value-accesors/text-value-accessor.directive.ts
+++ b/packages/elements-angular/src/directives/control-value-accesors/text-value-accessor.directive.ts
@@ -5,7 +5,7 @@ import { ValueAccessorDirective } from './value-accessor.directive';
 
 @Directive({
   selector:
-    'ino-autocomplete,ino-currency-input,ino-input,ino-markdown-editor,ino-textarea,ino-range,ino-select,ino-datepicker,ino-segment-group',
+    'ino-autocomplete,ino-currency-input,ino-input:not([type=number]):not([type=file]),ino-markdown-editor,ino-textarea,ino-select,ino-datepicker,ino-segment-group',
   providers: [
     {
       provide: NG_VALUE_ACCESSOR,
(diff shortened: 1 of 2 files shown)


packages/elements-angular/src/directives/control-value-accesors/text-value-accessor.directive.ts
```diff
@@ -5,7 +5,7 @@ import { ValueAccessorDirective } from './value-accessor.directive';
 
 @Directive({
   selector:
-    'ino-autocomplete,ino-currency-input,ino-input,ino-markdown-editor,ino-textarea,ino-range,ino-select,ino-datepicker,ino-segment-group',
+    'ino-autocomplete,ino-currency-input,ino-input:not([type=number]):not([type=file]),ino-markdown-editor,ino-textarea,ino-select,ino-datepicker,ino-segment-group',
```
BenPag, line 8: For using multiply `:not()` selectors I suggest writing it like here:
https://developer.mozilla.org/en-US/blog/css-not-pseudo-multiple-selectors/#negating_multiple_selectors

BenPag (changes requested):
Can you format this file (`numeric-value-accessor.directive.ts`) according the others. The `selector` value contains a whitespace.

Beside this the issue requires some test to prevent this issue in future. But I can´t see any 😄

packages/elements-angular/src/directives/control-value-accesors/text-value-accessor.directive.ts
janivo: I've just tried to add some simple jest test to the package but I think as long as we've got these conflicts with stencils outdated Jest version, I don't see us adding any new tests soon. I'll write that down in a ticket.

BenPag (approved):
Alright, then we should keep in mind that we write some tests later on.

PR #1218, opened 2024-02-01
ci: suppress NX unknown cache error

janivo:


BenPag (changes requested):
Can you add the env setting for the `yarn lint` and yarn `test` command too?

Also we check other workflow files.

BenPag (approved):

PR #1247, opened 2024-02-08
chore: generate release notes properly by using / separator

### LINKED ISSUE #1245: Update commit scope separator to be considered by our releases
### Description
The current commit scope separator used in our project, "|", is not recognized by NX release. This inconsistency prevents NX release from properly interpreting commit scopes, leading to potential issues during the release process. To address this, we need to change the commit scope separator to "/" to align with NX release requirements.

### Acceptance Criteria
- [ ] The commit scope separator in the project configuration is updated to "/".
- [ ] All team members are notified about the change and instructed to use "/" as the commit scope separator in their commit messages.
- [ ] NX release properly interprets commit scopes with the new separator "/" without any issues.

BenPag:
Closes #1245 

To validate these chages look here: https://github.com/BenPag/elements/actions/runs/7828956721

JFYI: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

## Proposed Changes

- use script to provide commit scopes and load them in workflow


commit-scope-enum.js
janivo, line 1: If we already have the `fs` reference in there anyway, then we could also read all components and all packages from the folder structure, couldn't we?

BenPag: In gerneral, yes we can. But we need some logic, that not all packages can be combined with the components.

E.g. we want to exclude: `ui-patterns/ino-icon` or `elements-react-example/ino-navbar`

BenPag: I add a possible solution

janivo (approved):

PR #1252, opened 2024-02-13
fix(elements/ino-nav-drawer): reinit listeners properly when switching between variants 

### LINKED ISSUE #1204: ino-drawer: Fix Behavior After Switching Between Docked and Mobile Variants
## Description
The current behavior of `ino-drawer` presents issues after switching from the docked to mobile variant. Clicking outside the drawer doesn't work as expected, and the listener appears to be still active after changing to the docked variant, toggling the nav drawer unexpectedly. This task involves addressing these issues for a more consistent and reliable user experience.

**Note: Tested with Angular**

## Acceptance Criteria:
- [ ] Clicking outside the `ino-drawer` in the mobile variant works as intended, toggling the drawer appropriately.
- [ ] The listener no longer bleeds after switching from the mobile variant, preventing unexpected toggling of the nav drawer.

BenPag:
Closes #1204 

## Proposed Changes

- reinitialize instance of MDCDrawer on variant change

### DIFF
--- a/packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
+++ b/packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
@@ -38,6 +38,7 @@ export class NavDrawer implements ComponentInterface {
    */
   private drawerInstance: MDCDrawer;
   private drawerEl: HTMLElement;
+  private reinitialize: boolean;
   @Element() el!: HTMLInoNavDrawerElement;
 
   /**
@@ -63,6 +64,16 @@ export class NavDrawer implements ComponentInterface {
    */
   @Prop() variant?: NavDrawerVariant = 'docked';
 
+  @Watch('variant')
+  variantChanged(newVariant: NavDrawerVariant) {
+    if (newVariant === 'mobile') {
+      this.activateMobileMode();
+    } else {
+      this.deactivateMobileMode();
+    }
+    this.reinitialize = true;
+  }
+
   /**
    * The aria-labels used for content and footer nav elements.
    * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role.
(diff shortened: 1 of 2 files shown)


packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
```diff
@@ -73,23 +84,34 @@ export class NavDrawer implements ComponentInterface {
...
-    // set initial value of open state
-    if (this.drawerInstance) {
-      this.drawerInstance.open = this.open || false;
+  componentDidUpdate() {
+    if (this.reinitialize) {
+      this.drawerInstance?.destroy();
+      this.initDrawerInstance();
+      this.reinitialize = false;
     }
+  }
```
janivo, line 111: Why do we need the `reinitialize` flag? Can't we just reinitialize whenever the `variant` changes (so in `variantChanged`)?

janivo (changes requested):

packages/elements/src/components/ino-nav-drawer/ino-nav-drawer.tsx
BenPag: We need this, because the drawerInstance needs to initialize after the ui updates. The watcher is executed before rerendering. 

I introduced this flag, because it is not necessary to reinit the drawerInstance by every component update.

janivo: Ah, I see. Can we name it a bit more explicit? Like `reinitializeAfterVariantChange`? Also please add a code comment on why we need this flag so we dont get confused in the future.

janivo (changes requested):
Ah, I see. Can we name it a bit more explicit? Like reinitializeAfterVariantChange? Also please add a code comment on why we need this flag so we dont get confused in the future.

janivo (approved):

PR #1270, opened 2024-02-21
test: fix flakiness of some tests

janivo:
- attempt to fix flakiness of the ino-button and ino-accordion test
- prevent retries as we don't want flaky tests in our main branch

### DIFF
--- a/packages/storybook/src/stories/ino-button/ino-button.spec.ts
+++ b/packages/storybook/src/stories/ino-button/ino-button.spec.ts
@@ -21,6 +21,7 @@ test.describe('ino-button', () => {
     // disable loading and wait for spinner to disappear
     await setAttribute(inoButton, 'loading', 'false');
     await expect(spinnerEl).toBeHidden();
+    await expect(inoButton).toHaveText(/Label/i);
 
     const { height: newHeight, width: newWidth } =
       await inoButton.boundingBox();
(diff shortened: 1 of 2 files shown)


packages/storybook/src/stories/ino-button/ino-button.spec.ts
```diff
@@ -1,32 +1,41 @@
 import { expect, Locator, test } from '@playwright/test';
-import { goToStory } from '../test-utils';
+import { goToStory, setAttribute } from '../test-utils';
 
 test.describe('ino-button', () => {
   let inoButton: Locator;
 
+  async function getButtonDimensions() {
```
BenPag, line 7: I found out that playwright has a `boundingBox()` function which can be called on an element direktly.
https://playwright.dev/docs/api/class-elementhandle#element-handle-bounding-box

For instance: 
```
// height and width are numbers
const { height, width } = await inoButton.getByRole('button').boundingBox();
```

packages/storybook/playwright.config.ts
```diff
@@ -29,7 +29,7 @@ export default defineConfig({
     trace: 'on-first-retry',
   },
   expect: {
-    timeout: process.env.CI ? 10000 : 5000,
+    timeout: process.env.CI ? 30000 : 10000,
```
BenPag, line 32: In my point of view 30s is really long. Do you have some evidence about this number?

BenPag (changes requested):

packages/storybook/playwright.config.ts
janivo: No thats just based on experience in another project

BenPag (approved):

PR #1289, opened 2024-03-08
deps: upgrade typescript to v5

janivo:
## Proposed Changes

- upgrade to Angular v17
- upgrade typescript to v5


BenPag (approved):

BenPag (changes requested):
on hold because of braking changes

PR #1325, opened 2024-03-15
ci: deploy storybook in release workflow

### LINKED ISSUE #1277: Merge Storybook Deploy Pipeline into Release Pipeline
### Acceptance Criteria
- [ ] Add manuel trigger to deploy storybook pipeline
- [ ] Merge steps in storybook pipeline into one single step so we don't have any depenedencies on uploaded artifacts
- [ ] Add input to skip generation of new versions

BenPag:
Closes #1277

Because we checkout the `pages` branch, we are not able to build the storybook directly. But I align the deployment of the storybook with the landingpage deployment.

**To be discussed**
Now I need a version as input for the deployment. But we can omit this.

To do this we can read the latest version from the root `package.json`. Same as now in the release workflow (see changes). During this it is not possible to deploy an other version than the latest one, but I do not know if this is needed.

## Proposed Changes

- provide manual storybook deployment
- deploy storybook with release workflow

### DIFF
--- a/.github/workflows/deploy-release.yml
+++ b/.github/workflows/deploy-release.yml
@@ -64,7 +64,7 @@ jobs:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
       - name: Get current package version
         id: package-version
-        uses: martinbeentjes/npm-get-version-action@v1
+        uses: martinbeentjes/npm-get-version-action@v1.3.1
 
   ShowVersion:
     needs: release
--- a/.github/workflows/deploy-release.yml
+++ b/.github/workflows/deploy-release.yml
@@ -78,4 +78,4 @@ jobs:
     if: ${{ inputs.configuration == 'release' }}
     uses: ./.github/workflows/deploy-storybook.yml
     with:
-     version: ${{ needs.release.outputs.version }}
+      version: ${{ needs.release.outputs.version }}
--- a/.github/workflows/deploy-release.yml
+++ b/.github/workflows/deploy-release.yml
@@ -28,6 +28,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write # needed for GitHub releases
+    outputs:
+      version: v${{ steps.package-version.outputs.current-version }}
     if: github.ref == 'refs/heads/master' || inputs.configuration == 'dryRun'
     steps:
       - name: Checkout repository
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-release.yml
```diff
@@ -60,3 +62,20 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: Get current package version
+        id: package-version
+        uses: martinbeentjes/npm-get-version-action@v1.3.1
+
+  ShowVersion:
```
BenPag, line 69: Currently for debug and testing purpose

.github/workflows/deploy-storybook.yml
```diff
@@ -1,24 +1,35 @@
...
-        uses: ./.github/workflows/artifacts-upload
+      - name: Upload Storybook artifact ⬆️
+        uses: actions/upload-artifact@v4
         with:
-          upload-storybook-artifact: true
+          name: storybook
+          path: packages/storybook/dist/
 
   deploy:
     needs: build
```
janivo, line 35: Can we just combine these steps? `build` & `deploy`?

.github/workflows/deploy-storybook.yml
```diff
@@ -1,24 +1,35 @@
...
+      version:
+        description: Latest elements version in format v[0-9].[0-9].[0-9]
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Latest elements version in format v[0-9].[0-9].[0-9]
+        required: true
+        type: string
```
janivo, line 15: When manually triggering the workflow, it would be great to just do it without providing a version in case you just want to fix some storybook-related stuff.

I expect the workflow to work roughly like this:
```
- install deps
- build storybook
- update symlink (only if a version was provided)
- update hosted versions (only if a version was provided)
- Deploy storybook (Commit message = version provided ? 'Deploy version xyz' : 'Update Storybook')
```

janivo (changes requested):

.github/workflows/deploy-storybook.yml
janivo: Ah, i saw your comment, makes sense

janivo (approved):

PR #1326, opened 2024-03-18
ci: add step in QC workflow to check uncommitted changes

### LINKED ISSUE #1314: Check for Uncommitted Changes After Build in Pipeline
### Description
To maintain code integrity and ensure that our committed changes include all generated files, we need to implement a check for uncommitted changes after the build process in our pipeline. This check will use the "has-changes" GitHub Action to detect any uncommitted changes in the repository and exit the pipeline with a failure status if changes are found. 

This is necessary because some info will be removed at random when using Stencil's dev mode (e.g.: documentation of CSS variables). To prevent this we want to ensure that our committed changes contain all generated files

### Acceptance Criteria
- [ ] The pipeline configuration includes an extra step after the build step has finished
- [ ] The extra steps correctly detects uncommitted changes that were generated by the `build` step, indicating that some generated files were not updated correctly in the local environment
- [ ] If uncommitted changes are found, the pipeline fails with an appropriate error message.
- [ ] The pipeline successfully completes without errors if there are no uncommitted changes.

### Further informations
- could come in handy: https://github.com/marketplace/actions/has-changes

BenPag:
Closes #1314 

## Proposed Changes

- add new step with usgae of `NathanielHill/fail-if-changes@latest`

Failing run: https://github.com/inovex/elements/actions/runs/8325292945/job/22778616258


.github/workflows/test.yml
```diff
@@ -46,6 +46,8 @@ jobs:
         run: yarn build
         env:
           NX_REJECT_UNKNOWN_LOCAL_CACHE: 0
+      - name: Check if there are changes
+        uses: NathanielHill/fail-if-changes@master
```
janivo, line 50: Should be use `@latest` instead of `@master`? It would at least help in terms of clarification because our main branch has the same name and I was wondering if you were comparing anything to the `master` branch 😅

BenPag: `@latest` is not working but using a specifc version works 👍

janivo (approved):

Issue #1290, opened 2024-03-08
Integrate NX executors in all packages


janivo:
### Description
To streamline our development process and leverage the benefits of the NX workspace, we need to integrate NX executors in all packages, including those built with React, Next and Vue frameworks. NX executors provide a standardized way to perform tasks such as building, testing, and linting across different types of projects within our workspace, ensuring consistency and efficiency in our development workflow.

### Acceptance Criteria
- [ ] NX executors are successfully integrated into all packages within the NX workspace, including React, Next and Vue projects.
- [ ] Executing NX tasks such as building, testing, linting, and serving works as expected
- [ ] Testing confirms that the integration of NX executors does not introduce any regressions or issues in the build and development processes

### Additional information
- consider a separate PR for each package

PR #1339, opened 2024-03-25
chore(landingpage): integrate nx executors for next.js

TobiasHeimGalindo:
Closes landingpage package in #1290


packages/landingpage/project.json
```diff
@@ -6,28 +6,36 @@
...
-        "prod": {
-          "command": "next start -p 4600"
-        }
+        "parallel": false,
+        "commands": [
+          "nx run landingpage:build",
+          "serve {projectRoot}/out"
+        ]
       }
     },
```
janivo, line 40: We are doing the same now in the storybook (building the static files & serving them). See here:
https://github.com/inovex/elements/blob/c04dbc0dba494b09bfb8591919b2fd0e4a4c9464/packages/storybook/project.json#L42-L48

Might be better to do it like this because there is build-in support from NX.

janivo (changes requested):

janivo (approved):

PR #1346, opened 2024-03-27
chore: streamline eslint configuration

janivo:
- fixes configuration problem within `packages/elements`
- streamlines eslint config in all packages
- add `lint:ci` script to prevent spam of ci output
- fix linting problems


TobiasHeimGalindo (approved):
Linting works well for me now :)

